PR #7, opened 2026-04-15
fix: advertise cPhyEnhance only when 1394a enhancements are enabled

gly11:
## Summary

- Derive the local Config ROM node capabilities from active controller state instead of a fixed constant
- Advertise `cPhyEnhance` only when 1394a PHY/link enhancements were actually enabled successfully
- Add host tests covering both the enabled and disabled capability cases

## Why

The driver currently serializes a fixed node capability value into the local Config ROM even when 1394a enhancements are skipped or fail to initialize.

Peers see `cPhyEnhance` advertised even though the controller did not actually enable the corresponding enhancement path. In that fallback case, the Config ROM no longer matches the driver's effective runtime state.

This change keeps the advertised local node capabilities aligned with the hardware state that was successfully brought up.

## What changed

- Replaced the fixed `kDefaultNodeCapabilities` constant with `MakeNodeCapabilities(bool phyEnhanceEnabled)` in `OHCIConstants.hpp`
- Updated `StageConfigROM()` to pass `phyConfigOk_` into the helper
- Added 2 tests: disabled and enabled `cPhyEnhance` cases

## Scope

Only changes local Config ROM capability advertisement.

Does **not** change:
- Bus-management policy
- The successful 1394a-enhanced path beyond making the advertisement conditional

## Test Plan

- [x] `cmake -S tests -B build/tests_build && cmake --build build/tests_build --target ASFWConfigROMTests`
- [x] `./build/tests_build/ASFWConfigROMTests --gtest_filter="ConfigROMBuilderTests.NodeCapabilities*"` — 2/2 passed
- [x] `./build.sh` — full Xcode build succeeded (0 errors, no new warnings)

### DIFF
--- a/ASFWDriver/Controller/ControllerCoreLifecycle.cpp
+++ b/ASFWDriver/Controller/ControllerCoreLifecycle.cpp
@@ -686,7 +686,7 @@ kern_return_t ControllerCore::StageConfigROM(uint32_t busOptions, uint32_t guidH
         (static_cast<uint64_t>(guidHi) << 32) | static_cast<uint64_t>(guidLo);
     const uint64_t effectiveGuid = (config_.localGuid != 0) ? config_.localGuid : hardwareGuid;
 
-    builder->Build(busOptions, effectiveGuid, ASFW::Driver::kDefaultNodeCapabilities,
+    builder->Build(busOptions, effectiveGuid, ASFW::Driver::MakeNodeCapabilities(phyConfigOk_),
                    config_.vendor.vendorName);
     if (builder->QuadletCount() < 5) {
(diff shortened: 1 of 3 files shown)


mrmidi:
Looks good for me. I've didn't even bothered with phyEnhanceEnabled (for mid 2000 devices is always enabled I guess). Just curious what kind of bizarre hardware do you working with :)

gly11:
> Looks good for me. I've didn't even bothered with phyEnhanceEnabled (for mid 2000 devices is always enabled I guess). Just curious what kind of bizarre hardware do you working with :)

Hi mrmidi, I’ve actually been working with some vintage film scanners that use FireWire. Not exactly the original use case for this project, but the low-level stuff maps surprisingly well. Still figuring out some quirks on my side 😄

PR #8, opened 2026-04-15
fix: widen async transaction generation to 16-bit

gly11:
## Summary
- Widen `PacketContext::generation` and `TransactionContext::generation` from `uint8_t` to `uint16_t`
- Add `static_assert` guards to prevent future narrowing
- The internal generation tracker uses 16-bit values for wrap-safe uniqueness; truncating to `uint8_t` silently discards the high byte, causing incorrect response matching across bus-reset cycles

Depends on #7

## Test plan
- [x] Existing C++ unit tests pass (`./build.sh --test-only`) — 67 async-related tests all green
- [x] `static_assert` compiles cleanly with C++23

### DIFF
--- a/ASFWDriver/Async/AsyncTypes.hpp
+++ b/ASFWDriver/Async/AsyncTypes.hpp
@@ -5,6 +5,7 @@
 #include <string>
 #include <functional>
 #include <span>
+#include <type_traits>
 #include <DriverKit/IOReturn.h>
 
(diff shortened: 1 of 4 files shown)


gly11:
Superseded by #9.

I folded the 16-bit generation fix into the new async/bus foundation PR so the related transaction-generation and completion-handling fixes can be reviewed together.

PR #12, opened 2026-05-01
fix(dice): bind Alesis MultiMix active stream profile

boggspa:
Recognize the exact Alesis MultiMix FireWire DICE identity and use a known active stream profile for its observed 12-in/2-out 48 kHz shape. Ignore inactive DICE streams with iso=-1 when deriving runtime caps and choose the discovered/profile iso channels during duplex restart instead of hardcoded defaults.

Tests: DeviceProtocolFactoryTests, DICETcatProtocolTests, DICEDuplexBringupControllerTests, DiceDuplexRestartCoordinatorTests.


mrmidi:
Thanks! If it works on real hardware — it's a perfect example how to add support to new DICE-based interface for current driver architecture.

PR #13, opened 2026-05-01
diag(dice): log TCAT runtime-cap discovery

boggspa:
Focused DICE diagnostics slice.

Adds logging around:
- DICE section-table reads
- global state reads
- TX/RX stream config reads
- cached runtime caps
- zero/partial runtime-cap failures

This is diagnostics-only:
- no binding changes
- no stream geometry changes
- no Midas support claim
- no UI or packaging changes

Tests:
- DICETcatProtocolTests


mrmidi:
Thanks for contributing! I hope to see your commits again :)

PR #18, opened 2026-05-27
fix(foundation): stabilize async discovery and bus reset recovery

gly11:
## Summary

This PR extracts the async, discovery, and bus-reset foundation fixes from the larger bring-up branch so they can be reviewed independently before higher-level protocol work.

### Async TX/RX Foundations

- Fix packet helper behavior used by async request/response routing
- Harden packet routing for request handlers and response sender paths
- Preserve response matching correctness across full generation values
- Retain full transaction response payloads instead of truncating larger results

### Bus Reset & Controller Bring-Up

- Add bounded recovery for manual resets that do not produce a bus-reset IRQ
- Delay discovery after accepted topology using the Apple-style scan delay
- Track reset diagnostics needed by recovery decisions
- Improve gap-count policy for unknown previous gap state and two-node local-root topologies
- Apply local contender eligibility while keeping root delegation policy-controlled

### Config ROM & Discovery

- Improve ROM scan fallback and completion behavior across speed retries
- Keep existing devices during inconclusive zero-ROM scans when topology still shows a remote link-active node
- Add discovery convergence checks to prevent premature device loss
- Extend device/unit discovery metadata and Swift parsing for the updated wire format

### Tests

- Async packet serialization compatibility tests
- Response sender header format tests
- Bus manager gap optimization tests
- Bus reset coordinator recovery tests
- Discovery convergence tests
- ROM scanner completion and node state machine tests
- Transaction storage tests

## Why this PR

These changes are shared foundations for subsequent protocol work, but they are useful and reviewable on their own. Keeping them separate reduces the follow-up PR size and lets reviewers focus first on async response handling, bus reset recovery, and discovery convergence.

This PR intentionally does not include higher-level protocol session/command plumbing, debug UI, local diagnostic scripts, install helpers, or documentation-only planning notes.

## Verification

```
cmake -S . -B build-pr1 -DCMAKE_BUILD_TYPE=Debug
cmake --build build-pr1 --target BusManagerGapOptimizationTests BusResetCoordinatorTests DiscoveryConvergenceTests ROMScanNodeStateMachineTests ROMScannerCompletionTests ResponseSenderHeaderFormatTests TransactionStorageTests ASFWPacketTests --parallel 8
```

Directly ran:

```
build-pr1/tests/BusManagerGapOptimizationTests
build-pr1/tests/BusResetCoordinatorTests
build-pr1/tests/DiscoveryConvergenceTests
build-pr1/tests/ROMScanNodeStateMachineTests
[...]

PR #19, opened 2026-05-27
feat(sbp2): add session registry and command flow

gly11:
## Summary

This PR builds on the foundation changes merged in #18 and splits out the SBP-2 session and command core from the larger SBP-2 bring-up branch. The branch has been rebased onto current `main`, so the visible diff is now limited to the SBP-2 session layer.

### SBP-2 Session Core
- Add login, reconnect, and logout session state handling, including bus-reset retry behavior, busy-timeout replay, and unsolicited status setup.
- Add a session registry for discovered SBP-2 units and command lifecycle management.
- Add command submission helpers for standard SCSI command flow and task management.
- Wire SBP-2 session operations through ControllerCore, DriverContext, and the user-client handler.

### ORB and Addressing
- Extend management and command ORB layout and status handling.
- Add full bus/node embedded-address handling and page-table/direct-descriptor updates.
- Add command wire formats for the user-client boundary.

### Tests
- Add login-session and session-registry coverage.
- Extend ORB and address-space tests for timeout, status, embedded-address, and command lifecycle behavior.

## Why this split

This PR intentionally excludes local debug UI, diagnostic handlers, install-helper changes, diagnostic scripts, and documentation experiments. Those can be reviewed separately or kept local. The async discovery and bus-reset foundations landed in #18; this layer adds SBP-2 session and command behavior on top of that.

## Verification

```
cmake -S . -B build-pr2 -DCMAKE_BUILD_TYPE=Debug
cmake --build build-pr2 --target AddressSpaceManagerTests SBP2ORBTests SBP2LoginSessionTests SBP2SessionRegistryTests --parallel 8
for t in AddressSpaceManagerTests SBP2ORBTests SBP2LoginSessionTests SBP2SessionRegistryTests; do "build-pr2/tests/$t" || exit $?; done
xcodebuild -project ASFW.xcodeproj -scheme ASFW -configuration Debug -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO build
```

After #18 merged, this branch was rebased onto current `upstream/main`; `git diff --check upstream/main..pr/sbp2-session-core` passes.

### DIFF
--- a/ASFWDriver/Protocols/SBP2/AddressSpaceManager.hpp
+++ b/ASFWDriver/Protocols/SBP2/AddressSpaceManager.hpp
@@ -7,6 +7,7 @@
 #include <cstring>
 #include <optional>
 #include <limits>
+#include <memory>
 #include <span>
 #include <unordered_map>
 #include <vector>
@@ -33,9 +34,20 @@ class AddressSpaceManager {
 public:
     // Callback invoked when a remote write arrives for a registered range.
     // Parameters: handle, offset within range, payload data.
+    //
+    // AddressSpaceManager invokes callbacks outside its lock. Callback targets
+    // must capture weak/shared lifetime state, not raw owning object pointers.
     using RemoteWriteCallback = std::function<void(uint64_t handle,
                                                    uint32_t offset,
                                                    std::span<const uint8_t> payload)>;
+    struct RemoteWriteCallbackSlot {
+        explicit RemoteWriteCallbackSlot(RemoteWriteCallback callbackIn)
+            : callback(std::move(callbackIn)) {}
+
+        RemoteWriteCallback callback;
+        std::atomic<bool> enabled{true};
+    };
+
     struct AddressRangeMeta {
         uint64_t handle{0};
         uint64_t address{0};
@@ -264,7 +276,7 @@ public:
             return Async::ResponseCode::DataError;
         }
 
-        RemoteWriteCallback callback;
+        std::shared_ptr<RemoteWriteCallbackSlot> callbackSlot;
         uint64_t handle = 0;
         uint32_t offset = 0;
 
@@ -293,14 +305,15 @@ public:
                 range->mappedBytes,
                 range->hasBacking ? 1u : 0u);
             WriteBytesLocked(*range, offset, payload);
-            callback = range->onRemoteWrite;
+            callbackSlot = range->remoteWriteSlot;
             handle = range->meta.handle;
             IOLockUnlock(lock_);
         }
 
         // Fire callback outside lock to avoid deadlock.
-        if (callback) {
-            callback(handle, offset, payload);
+        if (callbackSlot && callbackSlot->enabled.load(std::memory_order_acquire) &&
+            callbackSlot->callback) {
+            callbackSlot->callback(handle, offset, payload);
         }
 
         return Async::ResponseCode::Complete;
@@ -410,6 +423,10 @@ public:
 
     // Register a callback to fire when a remote write arrives for the given handle.
     // Must be called after AllocateAddressRange. Replaces any previous callback.
+    //
+    // Replacing or clearing the callback invalidates callbacks copied but not
+    // yet dispatched. Callback targets must still use weak/shared lifetime
+    // capture because callbacks run outside the manager lock.
     void SetRemoteWriteCallback(uint64_t handle, RemoteWriteCallback callback) {
         if (!lock_ || handle == 0) {
             return;
@@ -418,7 +435,7 @@ public:
         IOLockLock(lock_);
         auto it = ranges_.find(handle);
         if (it != ranges_.end()) {
-            it->second.onRemoteWrite = std::move(callback);
+            ReplaceRemoteWriteCallbackLocked(it->second, std::move(callback));
         }
         IOLockUnlock(lock_);
     }
@@ -460,7 +477,7 @@ private:
         AddressRangeMeta meta{};
         void* owner{nullptr};
         std::vector<uint8_t> buffer;
-        RemoteWriteCallback onRemoteWrite;
+        std::shared_ptr<RemoteWriteCallbackSlot> remoteWriteSlot;
         std::array<char, kDebugLabelCapacity> debugLabel{};
 
         OSSharedPtr<IOBufferMemoryDescriptor> descriptor{};
@@ -507,6 +524,18 @@ private:
 
     static const char* DebugLabelCString(const AddressRange& range) {
         return range.debugLabel[0] != '\0' ? range.debugLabel.data() : "unlabeled";
+    }
+
+    static void ReplaceRemoteWriteCallbackLocked(AddressRange& range,
+                                                 RemoteWriteCallback callback) {
+        if (range.remoteWriteSlot) {
+            range.remoteWriteSlot->enabled.store(false, std::memory_order_release);
+            range.remoteWriteSlot.reset();
+        }
+        if (callback) {
+            range.remoteWriteSlot =
+                std::make_shared<RemoteWriteCallbackSlot>(std::move(callback));
+        }
     }
 
     AddressRange* FindRangeByAddressLocked(uint64_t address, uint32_t length) {
--- a/ASFWDriver/Protocols/SBP2/AddressSpaceManager.hpp
+++ b/ASFWDriver/Protocols/SBP2/AddressSpaceManager.hpp
@@ -59,6 +59,7 @@ public:
     struct ReadSlice {
         uint64_t payloadDeviceAddress{0};
         uint32_t payloadLength{0};
+        std::shared_ptr<void> backingLease{};
     };
 
     explicit AddressSpaceManager(ASFW::Driver::HardwareInterface* hardware) noexcept
@@ -302,8 +303,8 @@ public:
                 static_cast<unsigned long long>(range->meta.address),
                 offset,
                 range->buffer.data(),
-                range->mappedBytes,
-                range->hasBacking ? 1u : 0u);
+                range->backing ? range->backing->mappedBytes : nullptr,
+                range->backing && range->backing->hasBacking ? 1u : 0u);
             WriteBytesLocked(*range, offset, payload);
             callbackSlot = range->remoteWriteSlot;
             handle = range->meta.handle;
@@ -334,13 +335,14 @@ public:
             return Async::ResponseCode::AddressError;
         }
 
-        if (!range->hasBacking || range->deviceAddress == 0) {
+        if (!range->backing || !range->backing->hasBacking ||
+            range->backing->deviceAddress == 0) {
             IOLockUnlock(lock_);
             return Async::ResponseCode::DataError;
         }
 
         const uint64_t offset = address - range->meta.address;
-        const uint64_t payloadAddress = range->deviceAddress + offset;
+        const uint64_t payloadAddress = range->backing->deviceAddress + offset;
         if (payloadAddress > 0xFFFF'FFFFULL) {
             IOLockUnlock(lock_);
             return Async::ResponseCode::DataError;
@@ -348,6 +350,7 @@ public:
 
         outSlice->payloadDeviceAddress = payloadAddress;
         outSlice->payloadLength = length;
+        outSlice->backingLease = range->backing;
 
         ASFW_ADDRSPACE_LOG(
             "AddressSpaceManager[%p] remote read-block label=%s addr=0x%012llx len=%u "
@@ -473,12 +476,27 @@ private:
     static constexpr uint64_t kAutoAddressAlignment = 8ULL;
     static constexpr std::size_t kDebugLabelCapacity = 64;
 
-    struct AddressRange {
-        AddressRangeMeta meta{};
-        void* owner{nullptr};
-        std::vector<uint8_t> buffer;
-        std::shared_ptr<RemoteWriteCallbackSlot> remoteWriteSlot;
-        std::array<char, kDebugLabelCapacity> debugLabel{};
+    struct AddressRangeBacking {
+        ~AddressRangeBacking() {
+            if (mapping) {
+                mapping->release();
+                mapping = nullptr;
+            }
+
+            if (dmaCommand) {
+                dmaCommand->CompleteDMA(kIODMACommandCompleteDMANoOptions);
+                dmaCommand.reset();
+            }
+
+            descriptor.reset();
+            mappedBytes = nullptr;
+            deviceAddress = 0;
+            hasBacking = false;
+        }
+
+        AddressRangeBacking() = default;
+        AddressRangeBacking(const AddressRangeBacking&) = delete;
+        AddressRangeBacking& operator=(const AddressRangeBacking&) = delete;
 
         OSSharedPtr<IOBufferMemoryDescriptor> descriptor{};
         OSSharedPtr<IODMACommand> dmaCommand{};
@@ -486,6 +504,15 @@ private:
         uint8_t* mappedBytes{nullptr};
         uint64_t deviceAddress{0};
         bool hasBacking{false};
+    };
+
+    struct AddressRange {
+        AddressRangeMeta meta{};
+        void* owner{nullptr};
+        std::vector<uint8_t> buffer;
+        std::shared_ptr<RemoteWriteCallbackSlot> remoteWriteSlot;
+        std::array<char, kDebugLabelCapacity> debugLabel{};
+        std::shared_ptr<AddressRangeBacking> backing{};
     };
 
     static uint64_t ComposeAddress(uint16_t hi, uint32_t lo) {
@@ -577,8 +604,8 @@ private:
                 range.owner,
                 static_cast<unsigned long long>(range.meta.address),
                 range.meta.length,
-                range.hasBacking ? 1u : 0u,
-                static_cast<unsigned>(range.deviceAddress));
+                range.backing && range.backing->hasBacking ? 1u : 0u,
+                static_cast<unsigned>(range.backing ? range.backing->deviceAddress : 0));
         }
     }
 
@@ -649,7 +676,7 @@ private:
 
         if (!dma.has_value()) {
 #ifdef ASFW_HOST_TEST
-            range.hasBacking = false;
+            range.backing.reset();
             return kIOReturnSuccess;
 #else
             ASFW_LOG(UserClient,
@@ -688,41 +715,30 @@ private:
         std::memset(mapped, 0, backingSize);
         OSSynchronizeIO();
 
-        range.descriptor = std::move(dma->descriptor);
-        range.dmaCommand = std::move(dma->dmaCommand);
-        range.mapping = mapping;
-        range.mappedBytes = mapped;
-        range.deviceAddress = dma->deviceAddress;
-        range.hasBacking = true;
+        auto backing = std::make_shared<AddressRangeBacking>();
+        backing->descriptor = std::move(dma->descriptor);
+        backing->dmaCommand = std::move(dma->dmaCommand);
+        backing->mapping = mapping;
+        backing->mappedBytes = mapped;
+        backing->deviceAddress = dma->deviceAddress;
+        backing->hasBacking = true;
+        range.backing = std::move(backing);
 
         return kIOReturnSuccess;
     }
 
     static void CleanupBacking(AddressRange& range) {
-        if (range.mapping) {
-            range.mapping->release();
-            range.mapping = nullptr;
-        }
-
-        if (range.dmaCommand) {
-            range.dmaCommand->CompleteDMA(kIODMACommandCompleteDMANoOptions);
-            range.dmaCommand.reset();
-        }
-
-        range.descriptor.reset();
-        range.mappedBytes = nullptr;
-        range.deviceAddress = 0;
-        range.hasBacking = false;
+        range.backing.reset();
     }
 
     static void SyncRange(AddressRange& range, uint64_t offset, uint64_t length) {
-        if (!range.hasBacking) {
+        if (!range.backing || !range.backing->hasBacking) {
             return;
         }
 
 #if defined(IODMACommand_Synchronize_ID)
-        if (range.dmaCommand) {
-            const kern_return_t syncKr = range.dmaCommand->Synchronize(
+        if (range.backing->dmaCommand) {
+            const kern_return_t syncKr = range.backing->dmaCommand->Synchronize(
                 0,
                 offset,
                 length);
@@ -770,14 +786,16 @@ private:
             data.size(),
             data.data(),
             range.buffer.data(),
-            range.mappedBytes,
+            range.backing ? range.backing->mappedBytes : nullptr,
             static_cast<unsigned long>(reinterpret_cast<uintptr_t>(data.data()) & 0x7ULL),
             static_cast<unsigned long>(reinterpret_cast<uintptr_t>(range.buffer.data()) & 0x7ULL),
-            static_cast<unsigned long>(reinterpret_cast<uintptr_t>(range.mappedBytes) & 0x7ULL));
+            static_cast<unsigned long>(
+                reinterpret_cast<uintptr_t>(range.backing ? range.backing->mappedBytes : nullptr) &
+                0x7ULL));
         CopyPayloadBytes(range.buffer.data() + static_cast<std::size_t>(offset), data);
 
-        if (range.hasBacking && range.mappedBytes) {
-            CopyPayloadBytes(range.mappedBytes + static_cast<std::size_t>(offset), data);
+        if (range.backing && range.backing->hasBacking && range.backing->mappedBytes) {
+            CopyPayloadBytes(range.backing->mappedBytes + static_cast<std::size_t>(offset), data);
             std::atomic_thread_fence(std::memory_order_release);
             SyncRange(range, offset, data.size());
         }
--- a/ASFWDriver/Protocols/SBP2/AddressSpaceManager.hpp
+++ b/ASFWDriver/Protocols/SBP2/AddressSpaceManager.hpp
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <algorithm>
+#include <array>
 #include <atomic>
 #include <cstdint>
 #include <cstring>
@@ -229,6 +230,9 @@ public:
         if (!lock_) {
             return kIOReturnBadArgument;
         }
+        if (data.size() > std::numeric_limits<uint32_t>::max()) {
+            return kIOReturnBadArgument;
+        }
 
         IOLockLock(lock_);
         auto it = ranges_.find(handle);
@@ -256,6 +260,9 @@ public:
         if (!lock_ || payload.empty()) {
             return Async::ResponseCode::AddressError;
         }
+        if (payload.size() > std::numeric_limits<uint32_t>::max()) {
+            return Async::ResponseCode::DataError;
+        }
 
         RemoteWriteCallback callback;
         uint64_t handle = 0;
@@ -275,7 +282,7 @@ public:
                 "AddressSpaceManager[%p] remote write label=%s addr=0x%012llx len=%zu src=%p "
                 "handle=0x%llx rangeAddr=0x%012llx off=%u buf=%p mapped=%p backing=%u",
                 this,
-                range->debugLabel,
+                DebugLabelCString(*range),
                 static_cast<unsigned long long>(address),
                 payload.size(),
                 payload.data(),
@@ -333,7 +340,7 @@ public:
             "AddressSpaceManager[%p] remote read-block label=%s addr=0x%012llx len=%u "
             "handle=0x%llx rangeAddr=0x%012llx off=%llu dma=0x%08x",
             this,
-            range->debugLabel,
+            DebugLabelCString(*range),
             static_cast<unsigned long long>(address),
             length,
             static_cast<unsigned long long>(range->meta.handle),
@@ -366,7 +373,7 @@ public:
             "AddressSpaceManager[%p] remote read-quadlet label=%s addr=0x%012llx "
             "handle=0x%llx rangeAddr=0x%012llx off=%u value=0x%08x",
             this,
-            range->debugLabel,
+            DebugLabelCString(*range),
             static_cast<unsigned long long>(address),
             static_cast<unsigned long long>(range->meta.handle),
             static_cast<unsigned long long>(range->meta.address),
@@ -424,7 +431,7 @@ public:
         IOLockLock(lock_);
         auto it = ranges_.find(handle);
         if (it != ranges_.end()) {
-            it->second.debugLabel = label != nullptr ? label : "unlabeled";
+            CopyDebugLabel(it->second, label);
         }
         IOLockUnlock(lock_);
     }
@@ -447,13 +454,14 @@ private:
     static constexpr uint32_t kAutoAddressWindowStartLo = 0x0010'0000u;
     static constexpr uint32_t kAutoAddressWindowEndLo = 0x0FFF'FFFFu;
     static constexpr uint64_t kAutoAddressAlignment = 8ULL;
+    static constexpr std::size_t kDebugLabelCapacity = 64;
 
     struct AddressRange {
         AddressRangeMeta meta{};
         void* owner{nullptr};
         std::vector<uint8_t> buffer;
         RemoteWriteCallback onRemoteWrite;
-        const char* debugLabel{"unlabeled"};
+        std::array<char, kDebugLabelCapacity> debugLabel{};
 
         OSSharedPtr<IOBufferMemoryDescriptor> descriptor{};
         OSSharedPtr<IODMACommand> dmaCommand{};
@@ -491,6 +499,16 @@ private:
         return end <= static_cast<uint64_t>(range.meta.length);
     }
 
+    static void CopyDebugLabel(AddressRange& range, const char* label) {
+        const char* source = label != nullptr ? label : "unlabeled";
+        range.debugLabel.fill('\0');
+        std::strncpy(range.debugLabel.data(), source, range.debugLabel.size() - 1);
+    }
+
+    static const char* DebugLabelCString(const AddressRange& range) {
+        return range.debugLabel[0] != '\0' ? range.debugLabel.data() : "unlabeled";
+    }
+
     AddressRange* FindRangeByAddressLocked(uint64_t address, uint32_t length) {
         const uint64_t end = address + static_cast<uint64_t>(length);
         if (end < address) {
@@ -525,7 +543,7 @@ private:
             ASFW_ADDRSPACE_LOG(
                 "AddressSpaceManager[%p] range label=%s handle=0x%llx owner=%p addr=0x%012llx len=%u backing=%u dma=0x%08x",
                 this,
-                range.debugLabel,
+                DebugLabelCString(range),
                 static_cast<unsigned long long>(range.meta.handle),
                 range.owner,
                 static_cast<unsigned long long>(range.meta.address),
@@ -557,6 +575,7 @@ private:
         }
 
         AddressRange range{};
+        CopyDebugLabel(range, nullptr);
         range.owner = owner;
         range.meta.handle = nextHandle_++;
         range.meta.address = start;
--- a/ASFWDriver/Protocols/SBP2/AddressSpaceManager.hpp
+++ b/ASFWDriver/Protocols/SBP2/AddressSpaceManager.hpp
@@ -2,10 +2,10 @@
 
 #include <algorithm>
 #include <atomic>
-#include <algorithm>
 #include <cstdint>
 #include <cstring>
 #include <optional>
+#include <limits>
 #include <span>
 #include <unordered_map>
 #include <vector>
@@ -272,9 +272,10 @@ public:
 
             offset = static_cast<uint32_t>(address - range->meta.address);
             ASFW_ADDRSPACE_LOG(
-                "AddressSpaceManager[%p] remote write addr=0x%012llx len=%zu src=%p "
+                "AddressSpaceManager[%p] remote write label=%s addr=0x%012llx len=%zu src=%p "
                 "handle=0x%llx rangeAddr=0x%012llx off=%u buf=%p mapped=%p backing=%u",
                 this,
+                range->debugLabel,
                 static_cast<unsigned long long>(address),
                 payload.size(),
                 payload.data(),
@@ -328,6 +329,18 @@ public:
         outSlice->payloadDeviceAddress = payloadAddress;
         outSlice->payloadLength = length;
 
+        ASFW_ADDRSPACE_LOG(
+            "AddressSpaceManager[%p] remote read-block label=%s addr=0x%012llx len=%u "
+            "handle=0x%llx rangeAddr=0x%012llx off=%llu dma=0x%08x",
+            this,
+            range->debugLabel,
+            static_cast<unsigned long long>(address),
+            length,
+            static_cast<unsigned long long>(range->meta.handle),
+            static_cast<unsigned long long>(range->meta.address),
+            static_cast<unsigned long long>(offset),
+            static_cast<unsigned>(payloadAddress));
+
         IOLockUnlock(lock_);
         return Async::ResponseCode::Complete;
     }
@@ -348,6 +361,17 @@ public:
         std::memcpy(outValue,
                     range->buffer.data() + static_cast<std::size_t>(offset),
                     sizeof(uint32_t));
+
+        ASFW_ADDRSPACE_LOG(
+            "AddressSpaceManager[%p] remote read-quadlet label=%s addr=0x%012llx "
+            "handle=0x%llx rangeAddr=0x%012llx off=%u value=0x%08x",
+            this,
+            range->debugLabel,
+            static_cast<unsigned long long>(address),
+            static_cast<unsigned long long>(range->meta.handle),
+            static_cast<unsigned long long>(range->meta.address),
+            offset,
+            *outValue);
 
         IOLockUnlock(lock_);
         return Async::ResponseCode::Complete;
@@ -392,6 +416,19 @@ public:
         IOLockUnlock(lock_);
     }
 
+    void SetDebugLabel(uint64_t handle, const char* label) {
+        if (!lock_ || handle == 0) {
+            return;
+        }
+
+        IOLockLock(lock_);
+        auto it = ranges_.find(handle);
+        if (it != ranges_.end()) {
+            it->second.debugLabel = label != nullptr ? label : "unlabeled";
+        }
+        IOLockUnlock(lock_);
+    }
+
     void ClearAll() {
         if (!lock_) {
             return;
@@ -416,6 +453,7 @@ private:
         void* owner{nullptr};
         std::vector<uint8_t> buffer;
         RemoteWriteCallback onRemoteWrite;
+        const char* debugLabel{"unlabeled"};
 
         OSSharedPtr<IOBufferMemoryDescriptor> descriptor{};
         OSSharedPtr<IODMACommand> dmaCommand{};
@@ -485,8 +523,9 @@ private:
         for (const auto& entry : ranges_) {
             const auto& range = entry.second;
             ASFW_ADDRSPACE_LOG(
-                "AddressSpaceManager[%p] range handle=0x%llx owner=%p addr=0x%012llx len=%u backing=%u dma=0x%08x",
+                "AddressSpaceManager[%p] range label=%s handle=0x%llx owner=%p addr=0x%012llx len=%u backing=%u dma=0x%08x",
                 this,
+                range.debugLabel,
                 static_cast<unsigned long long>(range.meta.handle),
                 range.owner,
                 static_cast<unsigned long long>(range.meta.address),
@@ -549,6 +588,7 @@ private:
 
     kern_return_t AllocateBacking(AddressRange& range) {
         const std::size_t size = static_cast<std::size_t>(range.meta.length);
+        const std::size_t backingSize = static_cast<std::size_t>(AlignUp(size, 16));
 
         // IOBufferMemoryDescriptor::Create expects memory-direction options only.
         // Cache policy is set at CreateMapping time, not in allocation options.
@@ -556,7 +596,7 @@ private:
 
         std::optional<ASFW::Driver::HardwareInterface::DMABuffer> dma;
         if (hardware_) {
-            dma = hardware_->AllocateDMA(size, options, 16);
+            dma = hardware_->AllocateDMA(backingSize, options, 16);
         }
 
         if (!dma.has_value()) {
@@ -576,7 +616,7 @@ private:
             kIOMemoryMapCacheModeInhibit,
             0,
             0,
-            size,
+            backingSize,
             0,
             &mapping);
         if (kr != kIOReturnSuccess || !mapping) {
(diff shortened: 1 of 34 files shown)


ASFWDriver/Protocols/SBP2/SBP2SessionRegistry.cpp
```diff
@@ -0,0 +1,668 @@
...
+            (void)record.session->Logout();
+        }
+        CleanupCommandResources(record);
+        CleanupManagementResources(record);
+    }
+    sessions_.clear();
+
+    if (lock_ != nullptr) {
+        IOLockFree(lock_);
+        lock_ = nullptr;
```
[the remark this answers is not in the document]
gly11: Fixed in the current branch head (166b8a7). The registry destructor now ends the lock guard scope before freeing lock_, so the guard cannot unlock a freed IOLock during shutdown.

ASFWDriver/Protocols/SBP2/SBP2LoginSession.cpp
```diff
@@ -0,0 +1,1719 @@
...
+    const kern_return_t prepareKr = orb->PrepareForExecution(localNode, speed, maxPayloadLog);
+    if (prepareKr != kIOReturnSuccess) {
+        ASFW_LOG(SBP2, "SBP2LoginSession::SubmitORB: PrepareForExecution failed: 0x%08x",
+                 prepareKr);
+        return false;
+    }
+
+    orb->SetFetchAgentWriteRetries(testFetchAgentWriteRetries_);
+    orb->SetAppended(true);
+    outstandingORBs_[MakeORBKey(orb->GetORBAddress())] = orb;
```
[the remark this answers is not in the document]
gly11: Fixed in the current branch head (166b8a7). Submitted command ORBs now start their timeout once the fetch-agent write succeeds and the target can fetch the ORB; normal completion and failure paths still cancel the timer.

ASFWDriver/Protocols/SBP2/SBP2SessionRegistry.cpp
```diff
@@ -0,0 +1,668 @@
...
+        rec->state.lastError = static_cast<int32_t>(result.transportStatus);
+        if (result.transportStatus == 0 && result.sbpStatus == Wire::SBPStatus::kNoAdditionalInfo) {
+            rec->state.lastError = 0;
+        }
+        rec->pendingCommandResult = std::move(result);
+        rec->activeCommandRequest.reset();
+        CleanupCommandResources(*rec);
+    });
+
+    if (!record->session->SubmitORB(orb.get())) {
```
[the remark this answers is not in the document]
gly11: Fixed in the current branch head (166b8a7). SubmitCommand now prepares and records the command under the registry lock, releases that lock before calling SubmitORB, and cleans up safely if submission fails. This avoids synchronous submit-failure callbacks re-entering the registry lock.

gly11:
The failing check on this PR is from the C++ coverage processing step, not from the SBP-2 code or tests. The run completed all 480 tests successfully, then failed while merging LLVM profile data.

I split the CI fix into #20. That PR isolates LLVM profile outputs for both test discovery and test execution, and its Build and Test check passes. Once #20 lands, this PR should only need a rerun/rebase against the updated workflow.

ASFWDriver/Protocols/SBP2/SBP2WireFormats.hpp
mrmidi: This is odd btw. 

both Apple and Linux indicates status byte as

```
Source   = (details >> 6) & 0x3
Response = (details >> 4) & 0x3
DeadBit  = (details >> 3) & 0x1
Length   = details & 0x7
```

worth checking

mrmidi: Ideally should be tested.

ASFWDriver/Protocols/SBP2/SBP2LoginSession.cpp
```diff
@@ -0,0 +1,1732 @@
...
+    if (status != Async::AsyncStatus::kSuccess) {
+        ASFW_LOG(SBP2, "SBP2LoginSession::OnReconnectWriteComplete: status=%s, retrying",
+                 Async::ToString(status));
+
+        SubmitDelayedCallback(100, [this]() { (void)Reconnect(); });
+        return;
+    }
+
+    // Reconnect ORB write ACK'd. Wait for status block from device.
+    ASFW_LOG(SBP2, "SBP2LoginSession: reconnect write ACK'd, waiting for status block");
```
mrmidi, line 814: My guess here should be something like this: https://github.com/apple-oss-distributions/IOFireWireSBP2/blob/875e5371555e095870675c925d3354d7cef222e7/Source/IOFireWireSBP2/IOFireWireSBP2Login.cpp#L2202

and not only log message

ASFWDriver/Protocols/SBP2/SBP2LoginSession.cpp
mrmidi, line 1: `CompleteLoginFromStatusBlock()`
`CompleteReconnectFromStatusBlock()`
`CompleteLogoutFromStatusBlock()`
`ProcessStatusBlock()`, which calls command completion as `cb(0, block.sbpStatus)`

all methods using `sbpStatus` which, as I mentioned in `SBP2WireFormats.hpp` comment may be parsed improperly. worth checking.

In general I'm trying to avoid 1.7k LOC files as you've may noticed, but let's leave it as is for now.

ASFWDriver/Protocols/SBP2/SBP2WireFormats.hpp
mrmidi, line 319: https://github.com/apple-oss-distributions/IOFireWireSBP2/blob/main/Source/IOFireWireSBP2/IOFireWireSBP2Login.h#L290-L303

Why diverges from Apple? Maybe you are right, I need to find the specs.

mrmidi: Confirmed. 
https://web.archive.org/web/20050118090648/http://www.t10.org/ftp/t10/drafts/sbp2/sbp2r04.pdf
§5.3.1 Request status, page 34:
```
0 No additional information to report
1 Request type not supported
2 Speed not supported
3 Page size not supported
4 Access denied
5 Logical unit not supported
6 Maximum payload too small
7 Reserved for future standardization
8 Resources unavailable
9 Function rejected
10 Login ID not recognized
11 Dummy ORB completed
12 Request aborted
FF16 Unspecified error
```

ASFWDriver/Protocols/SBP2/SBP2CommandORB.cpp
```diff
@@ -64,6 +66,10 @@ void SBP2CommandORB::DeallocateResources() noexcept {
 // ---------------------------------------------------------------------------
 
 void SBP2CommandORB::SetCommandBlock(std::span<const uint8_t> cdb) noexcept {
```
mrmidi: Silently truncates oversized CDBs:
`SetCommandBlock()` copies `min(cdb.size(), maxCommandBlockSize_)`, so a too-large SCSI CDB becomes a different command on the wire. The registry path currently prevents this here: https://github.com/mrmidi/ASFireWire/pull/19/changes#diff-a0a88c9ec126e3450b3893c72e5ef133889eff1855f2305ff0158f8f82cbe325R272 , so normal user-client flow is protected, but the class API itself is unsafe. Apple rejects this case: IOFireWireSBP2ORB::setCommandBlock() returns failure when length > maxCommandBlockSize here: https://github.com/apple-oss-distributions/IOFireWireSBP2/blob/875e5371555e095870675c925d3354d7cef222e7/Source/IOFireWireSBP2/IOFireWireSBP2ORB.cpp#L1579. I’d make return kern_return_t/bool, or at minimum refuse to overwrite the ORB when oversized.

Generally should be safe but worth guarding a bit here.

ASFWDriver/Protocols/SBP2/SBP2CommandORB.hpp
```diff
@@ -53,31 +54,36 @@ class SBP2CommandORB {
     void SetFlags(uint32_t flags) noexcept { flags_ = flags; }
     void SetMaxPayloadSize(uint16_t bytes) noexcept { maxPayloadSize_ = bytes; }
```
mrmidi: seems currently dead size, see https://github.com/mrmidi/ASFireWire/pull/19/changes#diff-b301c4a233c8cf32e6a2ea7e383bbefdba5d963d568e5590766d271ddf85705bR92

ASFWDriver/Protocols/SBP2/SBP2CommandORB.cpp
mrmidi, line 159: The SBP-2 command ORB max payload field is 4 bits. Current caller caps at 15 in https://github.com/mrmidi/ASFireWire/pull/19/changes#diff-cf28bdbf006a9e15485025a6f6989bd4e29b48029af89ea138dc01cc00334f8cR1271, so the path is okay, but `PrepareForExecution()` accepts a raw `uint16_t` and shifts it straight into options. I’d either assert/return error for > 15 or mask locally.

ASFWDriver/Protocols/SBP2/SBP2WireFormats.hpp
gly11: Fixed in c1c3ec5. The status block detail-byte accessors now match this layout exactly, and I added a packed detail-byte test covering Source/Response/Dead/Length decoding.

gly11: Fixed in c1c3ec5. I aligned the request status constants with SBP-2 §5.3.1 and the Apple SBP-2 headers, including logical-unit-not-supported, maximum-payload-too-small, resources-unavailable, dummy-ORB-completed, and request-aborted values. Added tests for the status-code constants.

ASFWDriver/Protocols/SBP2/SBP2PageTable.hpp
mrmidi, line 115: Page-table address ranges are leaked on command cleanup. 

`SBP2PageTable` allocates an address range, but it has no destructor calling Clear(). The registry later does `record.commandPageTable.reset()` here https://github.com/mrmidi/ASFireWire/pull/19/changes#diff-a0a88c9ec126e3450b3893c72e5ef133889eff1855f2305ff0158f8f82cbe325R665, which destroys the object but does not deallocate `pageTableHandle_`.

This only hits actual page-table commands, not the common direct-descriptor path. But any transfer that splits into multiple PTEs, e.g. `> 0xF000` bytes with current clipping, leaks an `AddressSpaceManager` range and its backing DMA memory until owner teardown. Repeated storage I/O could exhaust the SBP-2 address window / DMA allocations.

See https://github.com/apple-oss-distributions/IOFireWireSBP2/blob/59d58748a2b33ec35610d33c135bca8d29936300/Source/IOFireWireSBP2/IOFireWireSBP2ORB.cpp#L151 and https://github.com/apple-oss-distributions/IOFireWireSBP2/blob/59d58748a2b33ec35610d33c135bca8d29936300/Source/IOFireWireSBP2/IOFireWireSBP2ORB.cpp#L907

ASFWDriver/Protocols/SBP2/SBP2CommandORB.cpp
gly11: Fixed in 575a23f. PrepareForExecution() now rejects maxPayloadLog values greater than 15 with kIOReturnBadArgument, and I added a focused test for that guard.

ASFWDriver/Protocols/SBP2/SBP2CommandORB.cpp
gly11: Fixed in 575a23f. SetCommandBlock() now returns bool and rejects oversized CDBs without modifying the existing command block. The registry checks the return value, and there is a regression test covering the no-overwrite case.

ASFWDriver/Protocols/SBP2/SBP2CommandORB.hpp
gly11: Fixed in 575a23f. I removed the dead SBP2CommandORB maxPayloadSize_ state and SetMaxPayloadSize() setter, along with the registry call site. PrepareForExecution() continues to use the maxPayloadLog value derived by the login session submit path.

ASFWDriver/Protocols/SBP2/SBP2SessionRegistry.cpp
```diff
@@ -0,0 +1,685 @@
...
+    return flags;
+}
+
+SBP2TargetInfo BuildTargetInfoFromUnit(const Discovery::FWUnit& unit) {
+    SBP2TargetInfo info{};
+
+    info.managementAgentOffset = unit.GetManagementAgentOffset().value_or(0);
+    info.lun = static_cast<uint16_t>(unit.GetLUN().value_or(0) & 0xFFFF);
+
+    if (auto uc = unit.GetUnitCharacteristics(); uc.has_value()) {
```
mrmidi, line 52: `Unit_Characteristics` is decoded from the wrong bytes:

See: https://github.com/apple-oss-distributions/IOFireWireSBP2/blob/875e5371555e095870675c925d3354d7cef222e7/Source/IOFireWireSBP2/IOFireWireSBP2Login.cpp#L290

Impact: devices that advertise larger ORBs still get forced to `maxORBSize = 32`, so `maxCommandBlockSize = 12`. That rejects legitimate 16-byte CDBs at https://github.com/mrmidi/ASFireWire/pull/19/changes#diff-a0a88c9ec126e3450b3893c72e5ef133889eff1855f2305ff0158f8f82cbe325R272. Timeout can also become zero or wrong, depending on the value.

Linux: https://github.com/torvalds/linux/blob/eb3f4b7426cfd2b79d65b7d37155480b32259a11/drivers/firewire/sbp2.c#L1046

ASFWDriver/Protocols/SBP2/SBP2LoginSession.cpp
gly11: Fixed in 26bf1fa. On reconnect management-agent ACK, the session now starts a reconnect status timeout modeled after the Apple startReconnectTimer() path, using managementTimeoutMs + 1000ms. If no reconnect status block arrives, the existing reconnect timeout path falls back to a full login. The status-block path cancels the pending timeout, and the new test covers the ACK-without-status case.

ASFWDriver/Protocols/SBP2/SBP2LoginSession.cpp
gly11: Covered by c1c3ec5. The status block detail byte now decodes Source/Response/Dead/Length using the Apple/Linux bit layout, and the SBP-2 request-status constants were aligned with SBP-2 section 5.3.1. The login, reconnect, logout, and command-completion paths now consume the corrected StatusBlock fields, with tests for both detail-byte decoding and request-status constants.

ASFWDriver/Protocols/SBP2/SBP2PageTable.hpp
gly11: Fixed in 26d98f2. SBP2PageTable now clears its allocated page-table address range in its destructor, so resetting commandPageTable releases the AddressSpaceManager range and backing DMA memory. I also added a regression test that verifies a multi-PTE page table address is no longer readable after SBP2PageTable destruction.

ASFWDriver/Protocols/SBP2/SBP2SessionRegistry.cpp
gly11: Fixed in 81c6172. Unit_Characteristics now decodes the management timeout from bits [15:8] and the ORB size units from bits [7:0], matching the Apple implementation. The target info test now covers a 64-byte ORB advertisement, including the derived maxCommandBlockSize.

ASFWDriver/Protocols/SBP2/SBP2SessionRegistry.cpp
```diff
@@ -0,0 +1,685 @@
...
+    auto it = sessions_.find(handle);
+    if (it != sessions_.end()) {
+        CleanupCommandResources(it->second);
+        CleanupManagementResources(it->second);
+        sessions_.erase(it);
+    }
+    return true;
+}
+
+void SBP2SessionRegistry::ReleaseOwner(void* owner) {
```
mrmidi, line 535: Erases live sessions immediately after starting async logout.

Both Apple and linux is counting references for the object. `IOFireWireSBP2Login.cpp:979` etc. Probably worth exploring something more modern, than manual counting. `sbp2.c:1224`.

Probably something like converting `SBP2LoginSession` from `unique_ptr` ownership to `std::shared_ptr` and make callbacks capture `weak_ptr`, not raw `this`

ASFWDriver/Protocols/SBP2/SBP2SessionRegistry.cpp
```diff
@@ -0,0 +1,685 @@
...
+        sessions_.clear();
+    }
+
+    if (lock_ != nullptr) {
+        IOLockFree(lock_);
+        lock_ = nullptr;
+    }
+}
+
+std::expected<uint64_t, int> SBP2SessionRegistry::CreateSession(void* owner,
```
mrmidi, line 112: Registry allows duplicate sessions for the same owner/GUID/unit.

Apple prevents this at the user-client layer: the SBP-2 LUN is opened exclusively and `createLogin()` rejects a second login when fLogin already exists. Linux does not expose arbitrary session creation either; it has one sbp2_logical_unit per discovered LUN, stores a single login_id, and logs out before relogin. This registry currently keys only by generated handle, so the same owner can create multiple sessions for the same (guid, romOffset) and submit multiple exclusive logins. Since `BuildLoginORB()` always sets `kExclusiveLogin`, this can self-conflict on real hardware. I’d add a secondary key by (`guid`, `romOffset`) or (`owner`, `guid`, `romOffset`) and either return the existing handle or reject with `kIOReturnExclusiveAccess`.

My preference: key globally by (`guid`, `romOffset`) for active/logging-in/logged-in sessions.

refs:
`IOFireWireSBP2UserClient.cpp:1044`
`sbp2.c:965`
`sbp2.c:1180`
`IOFireWireSBP2LUN.cpp:425`
`sbp2.c:1180`

ASFWDriver/Protocols/SBP2/SBP2SessionRegistry.cpp
```diff
@@ -0,0 +1,685 @@
...
+        if (record.commandInFlight || record.commandORB) {
+            FailActiveCommandLocked(record,
+                                   static_cast<int>(kIOReturnAborted),
+                                   Wire::SBPStatus::kRequestAborted,
+                                   true);
+        }
+    }
+}
+
+void SBP2SessionRegistry::FailActiveCommandLocked(SBP2SessionRecord& record,
```
mrmidi, line 583: Fetch-agent failure path leaves `commandORB` retained after result is ready.

When session-level failure calls FailActiveCommandLocked(..., keepSessionTracking=false), the registry sets commandInFlight=false and stores a result, but intentionally does not reset commandORB. Then the ORB completion callback returns early because commandInFlight is already false at [SBP2SessionRegistry.cpp (line 329)](https://github.com/mrmidi/ASFireWire/pull/19/changes#diff-a0a88c9ec126e3450b3893c72e5ef133889eff1855f2305ff0158f8f82cbe325R329). The ORB address range is retained until the next command or session release.

ASFWDriver/Protocols/SBP2/SBP2SessionRegistry.cpp
```diff
@@ -0,0 +1,685 @@
...
+    state.lastError = record->state.lastError;
+    state.reconnectPending = (state.loginState == LoginState::Suspended);
+    return state;
+}
+
+bool SBP2SessionRegistry::SubmitInquiry(uint64_t handle, uint8_t allocationLength) {
+    return SubmitCommand(handle, SCSI::BuildInquiryRequest(allocationLength));
+}
+
+std::optional<std::vector<uint8_t>> SBP2SessionRegistry::GetInquiryResult(uint64_t handle) {
```
mrmidi: GetInquiryResult hides failed inquiry results

If INQUIRY completes with transport/SBP failure, `GetInquiryResult()` consumes the stored result and returns `nullopt`, which is indistinguishable from “not ready” to the caller. Generic `GetCommandResult()` preserves status, so maybe this helper is just convenience, but user-client `GetSBP2InquiryResult()` maps this to `kIOReturnNotFound`.

ASFWDriver/Protocols/SBP2/SBP2SessionRegistry.cpp
mrmidi, line 1: I suggest to add some tests here:

Add a `Unit_Characteristics` test with value = `0x00000408`: expect timeout `2000 ms`, max ORB `32`, max CDB `12`. Also add one with low byte advertising a bigger ORB, e.g. `0x00000410`: expect max ORB `64`, max CDB `44`.

Add a release-during-login/logout test using `DeferredFireWireBus`: start login/logout, release owner/session before completing the pending write, then complete the write.

ASFWDriver/Protocols/SBP2/AddressSpaceManager.hpp
mrmidi, line 323: Read-block response uses unpinned DMA backing after manager unlocks

`ResolveReadSlice()` returns only a raw DMA address/length, then DriverContext submits an AT read-block response using that address. If the owning range/session is released before OHCI completes the response DMA, `CleanupBacking()` can unmap memory that the AT response descriptor still references.

Apple avoids this by keeping readable SBP-2 memory behind an active pseudo address-space object with an owned descriptor:
- `IOFWSBP2PseudoAddressSpace.cpp:58`
- `IOFireWireSBP2Login.cpp:345`

Linux copies read-response bytes into request-owned storage and holds the request until async response transmit completion:
- `core-transaction.c:809`
- `core-transaction.c:853`
- `core-transaction.c:860`
- `core-transaction.c:653`
- `core-transaction.c:1126`

ASFW path:
- `AddressSpaceManager.hpp:302`
- `DriverContext.cpp:303`
- `ResponseSender.cpp:147`
- `DescriptorBuilder.cpp:508`

I’d either return a lease/guard from `ResolveReadSlice()` that is retained until AT-response completion, or copy the payload into response-owned DMA memory before submitting the response.

mrmidi, line 271: [P2] Remote-write callbacks can outlive callback clearing/release

ApplyRemoteWrite() copies range->onRemoteWrite while holding AddressSpaceManager's lock, unlocks, then invokes the copied callback. Calling outside the lock is good for avoiding deadlock, but it means SetRemoteWriteCallback(handle, {}) or owner/session teardown can run after the copy and before invocation. If the copied callback captures raw `this`, it can still call into an object that teardown believes is disconnected.

ASFW path:
- AddressSpaceManager.hpp:254
- AddressSpaceManager.hpp:289
- AddressSpaceManager.hpp:295
- AddressSpaceManager.hpp:406
- SBP2LoginSession.cpp:371
- SBP2LoginSession.cpp:394
- SBP2ManagementORB.cpp:197

Linux avoids this class with an address-handler unregister barrier. Address handlers are invoked under an RCU read-side section, and `fw_core_remove_address_handler()` removes the handler with `list_del_rcu()` then waits with `synchronize_rcu()`; the comment explicitly says that when removal returns the handler callback is no longer running on any CPU.
- core-transaction.c:912
- core-transaction.c:916
- core-transaction.c:614
- core-transaction.c:620
- core-transaction.c:622
- sbp2.c:973
- sbp2.c:975
- sbp2.c:1245
- sbp2.c:1247

Apple’s model also ties callbacks to registered address-space objects rather than copying callable targets out of a range. `IOFWPseudoAddressSpace::doWrite()` calls the registered writer/refcon, and SBP-2 tears down the status pseudo address space during login free. Address-space add/remove is serialized through the FireWire controller gate.
- IOFWPseudoAddressSpace.cpp:510
- IOFWPseudoAddressSpace.cpp:528
- IOFireWireSBP2Login.cpp:458
- IOFireWireSBP2Login.cpp:1738
- IOFireWireSBP2Login.cpp:1053
- IOFWAddressSpace.cpp:410
- IOFireWireController.cpp:5680
- IOFireWireController.cpp:5684

I’d either add an AddressSpaceManager-side lease/cancellation model with a “no callbacks running after clear/deallocate returns” guarantee, or require every callback target to use weak lifetime tokens. The SBP2LoginSession callback currently captures raw `this`; that one should become weak/self-token guarded at minimum.

ASFWDriver/Protocols/SBP2/AddressSpaceManager.hpp
```diff
@@ -392,6 +416,19 @@ class AddressSpaceManager {
         IOLockUnlock(lock_);
     }
 
+    void SetDebugLabel(uint64_t handle, const char* label) {
```
mrmidi, line 446: Current callers pass string literals, so no present bug. But `SetDebugLabel(const char*)` stores the pointer directly. If a future caller passes temporary storage, diagnostics can dereference stale memory. Either document “static string only” or store a fixed-size copied label/string.

ASFWDriver/Protocols/SBP2/AddressSpaceManager.hpp
mrmidi, line 239: WriteLocalData/ApplyRemoteWrite truncate huge span lengths before bounds check

Both paths cast `data.size()` / `payload.size()` to `uint32_t` before checking whether the write fits the range. If a caller ever passes a span larger than UINT32_MAX, the length can wrap and make `WithinRange()` / `FindRangeByAddressLocked()` validate the wrong size before `WriteBytesLocked()` copies the original span.

ASFW path:
- AddressSpaceManager.hpp:225
- AddressSpaceManager.hpp:244
- AddressSpaceManager.hpp:254
- AddressSpaceManager.hpp:266
- SBP2Handler.hpp:137
- SBP2Handler.hpp:146

Linux’s hardware path derives remote write/read block lengths from the FireWire packet data-length field, then allocates/copies using that bounded value:
- packet-header-definitions.h:83
- core-transaction.c:787
- core-transaction.c:790
- core-transaction.c:809

Linux cdev/user-facing send paths also cap user request lengths before allocation/submission:
- core-cdev.c:637
- core-cdev.c:638
- core-cdev.c:1541
- core-cdev.c:1542

Apple’s pseudo address-space APIs use `UInt32 len` consistently for registered address-space writes, so the low-level callback path is already width-limited rather than accepting a `size_t` span and narrowing later:
- IOFWAddressSpace.h:219
- IOFWPseudoAddressSpace.cpp:402
- IOFWPseudoAddressSpace.cpp:510
- IOFWSBP2PseudoAddressSpace.cpp:92
- IOFireWireSBP2Login.cpp:1738
- IOFireWireSBP2Login.cpp:1744

This is probably not reachable from normal FireWire packets, and the current SBP2 user-client path compares OSData length to a 32-bit scalar length before calling WriteLocalData. Still, since WriteLocalData() is a generic API and accepts `std::span`, I’d add an explicit guard before narrowing:

if (data.size() > std::numeric_limits<uint32_t>::max()) return kIOReturnBadArgument;
if (payload.size() > std::numeric_limits<uint32_t>::max()) return Async::ResponseCode::DataError;

ASFWDriver/UserClient/Handlers/SBP2Handler.hpp
```diff
@@ -128,10 +154,257 @@ class SBP2Handler {
...
+        if (!result.has_value()) {
+            return result.error();
+        }
+
+        args->scalarOutput[0] = *result;
+        args->scalarOutputCount = 1;
+        return kIOReturnSuccess;
+    }
+
+    kern_return_t StartSBP2Login(IOUserClientMethodArguments* args) {
```
mrmidi: Session operations do not validate the owning user-client

CreateSBP2Session() records `owner`, but later methods take only the numeric session handle and call registry methods that look up by handle alone. That means another user-client that guesses/learns a handle can start login, submit SCSI commands, fetch results, task-reset, or release someone else’s session. The handles are global and monotonic, so this is not a strong boundary.

ASFW path:
- SBP2Handler.hpp:164
- SBP2Handler.hpp:178
- SBP2Handler.hpp:188
- SBP2Handler.hpp:226
- SBP2Handler.hpp:263
- SBP2Handler.hpp:321
- SBP2Handler.hpp:366
- SBP2Handler.hpp:393
- SBP2SessionRegistry.cpp:111
- SBP2SessionRegistry.cpp:143
- SBP2SessionRegistry.cpp:246
- SBP2SessionRegistry.cpp:656

Apple validates login handles against the user-client’s own login object before acting:
- IOFireWireSBP2UserClient.cpp:1044
- IOFireWireSBP2UserClient.cpp:1102
- IOFireWireSBP2UserClient.cpp:1103
- IOFireWireSBP2UserClient.cpp:1187
- IOFireWireSBP2UserClient.cpp:1188

ASFWDriver/UserClient/Handlers/SBP2Handler.hpp
mrmidi, line 153: ReleaseOwner() releases address ranges before SBP-2 sessions

SBP2Handler::ReleaseOwner() calls AddressSpaceManager::ReleaseOwner(owner) before SBP2SessionRegistry::ReleaseOwner(owner). But command buffers and page tables are allocated with `record.owner`, so this can deallocate DMA-visible command backing before the registry cancels/cleans active ORBs.

ASFW path:
- SBP2Handler.hpp:153
- SBP2Handler.hpp:155
- SBP2Handler.hpp:158
- SBP2SessionRegistry.cpp:280
- SBP2SessionRegistry.cpp:302
- SBP2PageTable.hpp:111
- SBP2SessionRegistry.cpp:525
- SBP2SessionRegistry.cpp:533
- SBP2SessionRegistry.cpp:656
- SBP2SessionRegistry.cpp:661

Apple and Linux both tear down protocol/session state before freeing the associated address/callback resources:
- IOFireWireSBP2UserClient.cpp:407
- IOFireWireSBP2Login.cpp:1029
- IOFireWireSBP2Login.cpp:1053
- sbp2.c:1224
- sbp2.c:1241
- sbp2.c:1245
- sbp2.c:1247

mrmidi, line 146: GetSBP2SessionState() succeeds when no output buffer is available

This method requires five scalar outputs, but it only writes them if scalarOutput exists and scalarOutputCount >= 5. Otherwise it still returns kIOReturnSuccess with no state returned.

ASFW path:
- SBP2Handler.hpp:200
- SBP2Handler.hpp:204
- SBP2Handler.hpp:215
- SBP2Handler.hpp:223

Please reject missing/short output with kIOReturnBadArgument, like other scalar-output methods in this user-client layer.

ASFWDriver/UserClient/WireFormats/SBP2CommandWireFormats.hpp
mrmidi, line 1: This user-client ABI looks internally consistent, but I think it needs a little hardening before we rely on it.

Because cdbLength/outgoingLength/transferLength are client-controlled, SubmitSBP2Command() should reject unsupported sizes before copying into vectors or allocating command buffers. Today the CDB max check happens later in the registry, after request.cdb.assign(), and transferLength can later drive address-space/page-table allocation. I’d cap cdbLength early (Linux implements 16-byte CDBs; Apple rejects command blocks larger than the login’s max command block size), validate ToTarget outgoingLength == transferLength before copying, and add an explicit supported transferLength cap for this ABI path.

I’d also reject nonzero _reserved bytes and require captureSenseData to be 0/1 so future ABI flags do not get silently accepted by old drivers.

Finally, since these packed structs are shared ABI, please add static_asserts for size/offsets and a short comment that these are DriverKit user-client/native-endian records, not SBP-2 big-endian bus wire structures.

Refs:
SBP2CommandWireFormats.hpp:7
SBP2CommandWireFormats.hpp:8
SBP2CommandWireFormats.hpp:9
SBP2CommandWireFormats.hpp:10
SBP2CommandWireFormats.hpp:12
SBP2CommandWireFormats.hpp:13
SBP2CommandWireFormats.hpp:14
SBP2CommandWireFormats.hpp:18
SBP2Handler.hpp:283
SBP2Handler.hpp:313
SBP2Handler.hpp:315
SBP2SessionRegistry.cpp:272
SBP2SessionRegistry.cpp:280
SBP2SessionRegistry.cpp:302
Linux: firewire/sbp2.c:193
Linux: firewire/sbp2.c:197
Linux: firewire/sbp2.c:311
Linux: firewire/sbp2.c:1150
Apple: IOFireWireSBP2ORB.cpp:1579
Apple: IOFireWireSBP2ORB.cpp:1585

ASFWDriver/Service/DriverContext.cpp
```diff
@@ -44,12 +46,13 @@ void ServiceContext::Reset() {
     deps.configRomStager.reset();
     deps.interrupts.reset();
     deps.topology.reset();
-    deps.fcpResponseRouter.reset(); // Clean up FCP router
+    deps.fcpResponseRouter.reset();  // Clean up FCP router
     deps.sbp2AddressSpaceManager.reset();
```
mrmidi, line 69: These SBP2/FCP objects are destroyed before asyncSubsystem, but EnsureSbp2Deps() registers PacketRouter handlers that captured raw pointers to them.

If AR routing can still run during teardown, the PacketRouter callbacks can dereference freed sbp2Manager/fcpRouter/responder pointers. I’d stop/clear the async subsystem or PacketRouter handlers before releasing these dependencies, or make the handlers capture lifetime-safe tokens.

ASFWDriver/Protocols/SBP2/SBP2SessionRegistry.hpp
```diff
@@ -0,0 +1,143 @@
+#pragma once
+
+// SBP-2 Session Registry — bridges discovery metadata to SBP2LoginSession instances.
```
mrmidi, line 3: The header says sessions are owned/keyed by (guid, romOffset), but the implementation exposes and looks up sessions by handle only.

ASFWDriver/Service/DriverContext.cpp
```diff
@@ -137,116 +140,185 @@ void DriverWiring::EnsureDeps(ASFWDriver* driver, ::ServiceContext& ctx) {
...
+                        Async,
+                        "SBP2 AR write-quadlet req: src=0x%04x offset=0x%012llx len=4 -> rcode=0x%x",
+                        packet.sourceID,
+                        static_cast<unsigned long long>(destOffset),
+                        static_cast<unsigned>(result));
+                    return result;
+                });
+
+            // Block write requests (tCode 0x1): SBP-2 first, then FCP fallback.
+            router->RegisterRequestHandler(
```
mrmidi, line 224: This registers a second tCode 0x1 block-write handler after ASFWDriver::Start() already registered an FCP-only handler.

PacketRouter overwrites existing handlers, so the final behavior is probably the intended SBP2-first/FCP-fallback path, but startup routing now depends on registration order. I’d keep only this combined registration path.

ASFWDriver/Protocols/SBP2/SBP2SessionRegistry.cpp
gly11: Fixed in 8164308. SBP2LoginSession is now held by shared_ptr, session async callbacks capture weak_from_this(), and released sessions are retained in the registry retirement list until logout status or timeout completes.

ASFWDriver/Protocols/SBP2/SBP2SessionRegistry.cpp
gly11: Fixed in 7f9dc8b. CreateSession now rejects an existing target session keyed by guid and romOffset with kIOReturnExclusiveAccess, including sessions that are still retiring after logout was started.

ASFWDriver/Protocols/SBP2/SBP2SessionRegistry.cpp
gly11: Fixed in 138c7f1. Fetch-agent retry exhaustion now fails the submitted ORB directly, removes its ORB tracking, cancels its timeout, and releases the command ORB address range before reporting completion.

ASFWDriver/Protocols/SBP2/SBP2SessionRegistry.cpp
gly11: Fixed in d93376b. GetInquiryResult now returns the completed CommandResult even when transportStatus or sbpStatus reports failure, so a completed failed INQUIRY is no longer collapsed into nullopt / kIOReturnNotFound.

ASFWDriver/Protocols/SBP2/SBP2SessionRegistry.cpp
gly11: Fixed in 9bb9464. I added the requested Unit_Characteristics coverage for 0x00000408 and kept the existing 0x00000410 case. I also added DeferredFireWireBus release-edge tests covering ReleaseOwner while login write is pending and ReleaseSession while logout write is pending, with the logout case retained until status completion.

ASFWDriver/UserClient/Handlers/SBP2Handler.hpp
gly11: Fixed in 0213018. SBP2 session operations now validate the handle against the owning user-client before starting login, querying state, submitting inquiry/commands, fetching inquiry/command results, submitting task management, or releasing a session. The user-client dispatch passes its client object through to the SBP2 handler, and the registry test covers non-owner rejection for those paths.

ASFWDriver/UserClient/Handlers/SBP2Handler.hpp
gly11: Fixed in 00b109e. SBP2Handler::ReleaseOwner() now tears down SBP-2 sessions before releasing owner address ranges, so command buffers and page-table backing remain available while the registry cancels and cleans active ORBs.

gly11: Fixed in 1c96fe1. GetSBP2SessionState() now returns kIOReturnBadArgument when the scalar output buffer is missing or has fewer than the five required fields. I added a handler-level regression test for missing, short, and valid output buffers.

ASFWDriver/UserClient/WireFormats/SBP2CommandWireFormats.hpp
gly11: Fixed in 11ddc22. The SBP2 command user-client ABI now documents these records as native-endian DriverKit user-client records, has size/offset static_asserts, rejects nonzero reserved bytes and invalid captureSenseData values, caps CDB length at 16 bytes before copying, validates ToTarget outgoingLength against transferLength, rejects unexpected outgoing payloads for non-ToTarget commands, and applies an explicit 16 MiB transfer-length cap before registry allocation.

ASFWDriver/Protocols/SBP2/SBP2SessionRegistry.hpp
gly11: Fixed in 71e7b25. The header now describes the actual model: sessions are created for an owner and target pair, public operations use opaque handles with owner validation, and duplicate live targets are rejected by guid/romOffset.

ASFWDriver/Protocols/SBP2/AddressSpaceManager.hpp
gly11: Fixed in 98b0a90. ResolveReadSlice() now returns a backing lease with the DMA address and length. DriverContext passes that lease into ResponseSender, and ResponseSender retains it by the AT response tail descriptor until the AT response completion is drained; teardown also clears any outstanding response leases.

gly11: Fixed in 4d50ee4. WriteLocalData() and ApplyRemoteWrite() now reject spans larger than UINT32_MAX before narrowing to uint32_t, with AddressSpaceManagerTests covering both guards.

gly11: Fixed in c53fcd2. AddressSpaceManager now uses cancellable callback slots so clearing or replacing a callback invalidates copied callbacks that have not been dispatched yet. SBP-2 status callbacks also use weak/shared lifetime tokens: SBP2LoginSession captures weak_from_this(), and management ORBs capture their shared async state instead of raw object lifetime.

ASFWDriver/Service/DriverContext.cpp
gly11: Fixed in 5dcbd82. ASFWDriver::Start() no longer registers the early FCP-only tCode 0x1 handler; startup now uses the single combined SBP2-first/FCP-fallback registration in DriverWiring::EnsureSbp2Deps().

ASFWDriver/Protocols/SBP2/AddressSpaceManager.hpp
gly11: Fixed in 4d50ee4. SetDebugLabel() now copies labels into range-owned fixed-size storage, so diagnostics no longer depend on the caller pointer lifetime.

ASFWDriver/Service/DriverContext.cpp
gly11: Fixed in d7da945. Service teardown now stops the async subsystem and clears PacketRouter handlers before releasing the SBP2/FCP dependencies, so AR routing cannot call handlers that captured those objects after teardown begins.

mrmidi:
Thanks @gly11 — this looks like a great starting point for the SBP-2 implementation.

I’m going to merge this as a foundation. The next step will be to adapt and split the SBP-2 pieces so they align better with the newer protocol/device architecture we now have on the DICE branch. That follow-up restructuring can happen separately; this PR gives us a useful base to build from.

gly11:
I have a few follow-up draft branches from the old stack, but I do not want to open follow-up PRs against the wrong base or architecture.

Should small independent fixes still target main, or would you prefer follow-up work to target DICE while the newer protocol/device architecture is being developed there?

I can re-split the queue so SBP-2-related work waits for the DICE-aligned structure, while only truly independent app/debug fixes go to main if that is preferred.

mrmidi:
Preferably wait a bit. OR f you have some tokens to burn and time to debug/test — read below :).

I have some ideas how to re-organize different protocols: main idea could be seen in f19d2d2. Bierfly — decouple audio leakage from discovery and clear separation how protocols should be loaded. In the same time — i'm finishing full Bus/IRM manager implementation. So the goal is not to grow the driver for every single device, but make it hardware agnostic where possible — follow the specs first, quirks later.
Also in the DICE branch: some critical bug fixes regarding OHCI hw initialization, some IEEE 1394 constants fixes, etc. Generally it's much more reliable with my hardware right now. (audio performance degraded a bit, but that's not what I care right now). I've already ported your config rom fixes there too — please test if it works with your Nikon if you can.

So for SBP-2 i see it it like that (it's a draft, just the core idea):

```
ASFWDriver/
  DeviceProfiles/
    Storage/
      StorageProfileTypes.hpp
      StorageProfileRegistry.hpp
      StorageProfileRegistry.cpp
      Generic/
        GenericSBP2Profiles.hpp
        GenericSBP2Profiles.cpp
      Vendors/  <- probably not needed at all for SBP-2

  Storage/
    StorageCoordinator.hpp
    StorageCoordinator.cpp
    StorageRuntimeRegistry.hpp
    StorageRuntimeRegistry.cpp

    SCSI/ <- SCSIControllerDriverKit part
      DriverKitSCSIController.hpp
      DriverKitSCSIController.cpp
      DriverKitSCSIDevice.hpp
      DriverKitSCSIDevice.cpp
      DriverKitSCSIRequestBridge.hpp
      DriverKitSCSIRequestBridge.cpp

  Protocols/
    Storage/
      ISCSITransport.hpp
      SCSITransportTypes.hpp

      SBP2/
        SBP2StorageTransport.hpp
        SBP2StorageTransport.cpp

        Core/
          SBP2Constants.hpp
          SBP2Types.hpp
          SBP2TargetInfo.hpp
          SBP2TargetParser.hpp
          SBP2LoginSession.hpp
          SBP2LoginSession.cpp
          SBP2SessionState.hpp
          SBP2GenerationBinding.hpp

        ORB/
          SBP2WireFormats.hpp
          SBP2ManagementORB.hpp
          SBP2ManagementORB.cpp
          SBP2CommandORB.hpp
          SBP2CommandORB.cpp
          SBP2PageTable.hpp
          SBP2PageTable.cpp
          SBP2ORBExecutor.hpp
          SBP2ORBExecutor.cpp

        Addressing/
          SBP2AddressSpaceManager.hpp
          SBP2AddressSpaceManager.cpp
          SBP2AddressSpaceLease.hpp
          SBP2AddressSpaceLease.cpp
          SBP2StatusBlockAddress.hpp

        Status/
          SBP2StatusBlock.hpp
          SBP2StatusBlock.cpp
          SBP2StatusMapper.hpp
          SBP2StatusMapper.cpp
```          

Ping me on Discord - we could chat about it more there!

gly11:
> Preferably wait a bit. OR f you have some tokens to burn and time to debug/test — read below :).
> 
> I have some ideas how to re-organize different protocols: main idea could be seen in [f19d2d2](https://github.com/mrmidi/ASFireWire/commit/f19d2d21fd3660610fe535d21659e2c1473a4160). Bierfly — decouple audio leakage from discovery and clear separation how protocols should be loaded. In the same time — i'm finishing full Bus/IRM manager implementation. So the goal is not to grow the driver for every single device, but make it hardware agnostic where possible — follow the specs first, quirks later. Also in the DICE branch: some critical bug fixes regarding OHCI hw initialization, some IEEE 1394 constants fixes, etc. Generally it's much more reliable with my hardware right now. (audio performance degraded a bit, but that's not what I care right now). I've already ported your config rom fixes there too — please test if it works with your Nikon if you can.
> 
> So for SBP-2 i see it it like that (it's a draft, just the core idea):
> 
> ```
> ASFWDriver/
>   DeviceProfiles/
>     Storage/
>       StorageProfileTypes.hpp
>       StorageProfileRegistry.hpp
>       StorageProfileRegistry.cpp
>       Generic/
>         GenericSBP2Profiles.hpp
>         GenericSBP2Profiles.cpp
>       Vendors/  <- probably not needed at all for SBP-2
> 
>   Storage/
>     StorageCoordinator.hpp
>     StorageCoordinator.cpp
>     StorageRuntimeRegistry.hpp
>     StorageRuntimeRegistry.cpp
> 
>     SCSI/ <- SCSIControllerDriverKit part
>       DriverKitSCSIController.hpp
>       DriverKitSCSIController.cpp
>       DriverKitSCSIDevice.hpp
>       DriverKitSCSIDevice.cpp
>       DriverKitSCSIRequestBridge.hpp
>       DriverKitSCSIRequestBridge.cpp
> 
>   Protocols/
>     Storage/
>       ISCSITransport.hpp
>       SCSITransportTypes.hpp
> 
>       SBP2/
>         SBP2StorageTransport.hpp
>         SBP2StorageTransport.cpp
> 
>         Core/
>           SBP2Constants.hpp
>           SBP2Types.hpp
>           SBP2TargetInfo.hpp
>           SBP2TargetParser.hpp
>           SBP2LoginSession.hpp
>           SBP2LoginSession.cpp
>           SBP2SessionState.hpp
>           SBP2GenerationBinding.hpp
> 
>         ORB/
>           SBP2WireFormats.hpp
>           SBP2ManagementORB.hpp
>           SBP2ManagementORB.cpp
>           SBP2CommandORB.hpp
>           SBP2CommandORB.cpp
>           SBP2PageTable.hpp
>           SBP2PageTable.cpp
>           SBP2ORBExecutor.hpp
>           SBP2ORBExecutor.cpp
> 
>         Addressing/
>           SBP2AddressSpaceManager.hpp
>           SBP2AddressSpaceManager.cpp
>           SBP2AddressSpaceLease.hpp
>           SBP2AddressSpaceLease.cpp
>           SBP2StatusBlockAddress.hpp
> 
>         Status/
>           SBP2StatusBlock.hpp
>           SBP2StatusBlock.cpp
>           SBP2StatusMapper.hpp
>           SBP2StatusMapper.cpp
> ```
> 
> Ping me on Discord - we could chat about it more there!

That makes sense. I’ll test the DICE branch with my Nikon hardware next.

If I find any issues, I’ll report them in an issue or open a focused PR with a fix.

I’ll also keep an eye on Discord for follow-up discussion.

PR #21, opened 2026-06-04
feat(dice): recognize Midas Venice as a deferred fail-closed DICE device

boggspa:
## What

Recognizes the **Midas Venice** (F-series) FireWire interface in the audio DeviceProfiles single-source-of-truth, so the driver identifies it in discovery/diagnostics instead of treating it as an unknown device.

## Fail-closed by design

Recognition is **identity-only**: the audio profile reports integration mode `kNone` — the same "recognized but deferred" treatment DICE already gives the deferred Focusrite multistream models (SPro40 / SPro26 / Liquid 56). The factory does **not** instantiate a runtime protocol, so **no CoreAudio endpoint is published**.

This is deliberate. Standard DICE stream caps come back `runtime_caps_not_ready` on this hardware (0/0 channels) until a DICE **EAP / current-config** probing path and clock setup exist. Publishing a guessed endpoint would mislead testers and disturb hardware state — so this recognizes now, and publishes later.

## IDs / provenance

Observed on real Midas Venice hardware via FFADO + Config ROM: vendor OUI `0x10c73f`, model `0x000001` (FFADO `driver="DICE"`, `mixer="Generic_Dice_EAP"`). The match is vendor-wide so the F16 / F24 / F32 line is recognized under one honest "Venice" name rather than guessing the variant.

## Changes

- `AudioDeviceIds`: add `kMidasVendorId` / `kVeniceModelId` + display names
- `Vendors/MidasAudioProfiles`: identity match + `kNone` audio profile (mirrors the other vendor providers)
- `AudioProfileRegistry`: wire the Midas provider into both lookups
- `DeviceProtocolFactory`: re-export the constants for existing call sites
- tests: Midas coverage in `DeviceProtocolFactoryTests` + `AudioProfileRegistryTests`

## Testing

`cmake -S tests -B build/tests_build && ctest` → **849/849 pass** (including 2 new Midas tests). No behavior change for any existing device.

The EAP probing path — the actual next step toward Midas audio — is intentionally a separate future slice.


boggspa:
Closing this — superseded. The fail-closed Midas Venice recognition here has been overtaken by the full Venice dual-stream support (#32, completed in #40/#41). No longer needed. Thanks!

PR #22, opened 2026-06-04
docs(dice): Alesis MultiMix hardware field notes for the audio rewrite

boggspa:
## What

Adds a short reference doc capturing **device-validated DICE/TCAT findings** from bringing a real **Alesis MultiMix FireWire** from clicky/poppy capture to clean multi-hour Logic recording (12-in / 2-out @ 48 kHz).

Docs-only — no code changes.

## Why

The code these findings originally lived in is **superseded by the direct-only audio rewrite** — but the findings themselves are the kind of hard-won, hardware-specific knowledge that's easy to lose in a rewrite. This records them so they can inform the new RX path:

- resolve from DICE **section-table offsets**, not absolute `GLOBAL_*` addresses
- which DICE registers proved authoritative (active TX/RX stream tables, `sampleRate`, AVS DBS/system-mode, …)
- **RX startup-hold** for expected zero-fill; **dual rebase** (startup + transport lock)
- high-water **backlog slewing** before overflow; `SYT == 0xffff` = no-info, **not** clock loss
- **AM824 data-block-count continuity** as the authoritative discontinuity signal

The device identity / stream-shape findings are already upstream via the Alesis profile (#12); they're noted in the doc only for self-containment.


boggspa:
Closing this for hygiene — the Alesis MultiMix field notes here are stale against the current tree (pre-DICE-merge era). The observations are preserved in git history and can be re-submitted fresh if useful for the audio work later. Thanks!

PR #26, opened 2026-06-10
lab: lift the Saffire UpdateIsochBufferParams latency model

boggspa:
## What

The `UpdateIsochBufferParams` trace, lifted into code — *"the whole model is liftable now, no bench capture in the loop"*. Branches off your current DICE tip (includes your seqlock/dump/geometry work).

## The pieces

- **`Protocols/Audio/DICE/SaffireIsochLatency.hpp`** — header-only constexpr unit encoding the trace verbatim:
  - delay-packet table: `{14,2} / {16,6} / {18,10} / {20,14}`, **+2** at 88.2/96k, **+4** at 176.4/192k (both columns)
  - DMA-program depth `160/80/40` packets; frames-per-packet `8/16/32`
  - derived safety-offset helpers (`delayPackets × framesPerPacket`)
  - full provenance in the header: host-internal chain (latencyMode = host preference, not a register; writer called only from `initHardware`/`RestartStreaming`), DICE-quirk note, unsupported rates rejected
- **Tests** — every table cell pinned, bumps, depth, packet-time math, constexpr usability, rejection. Suite: **18,854 checks, 0 failures**.
- **Dext wiring** — your `SetOutputSafetyOffset(0)` placeholder now takes the model value at **mode kLow** (your trace's recommendation: start 0–1 for TX, widen only if `framesWithoutPacket` climbs — that counter is already in the dump). `SetOutputLatency` carries the same value for now — *which ADK property maps to which IOAudioFamily field* is a bench question; both start at the model value so the bench can observe. Lookup failure logs and falls back to defaults.

## Two prose/table discrepancies in the trace — please adjudicate

The code encodes the **table** in both cases:

1. *"safest mode = 160 frames (~3.3 ms)"* for the **output** offset at 48 kHz — but the table gives output mode 3 = 14×8 = **112** frames; 160 = the **input** column (20×8).
2. *"a constant ~20 ms hardware buffer regardless of rate"* — in packet time 160/80/40 × 125 µs = **20/10/5 ms**; only the ≤48k family is 20 ms. (If the intent was constant *frames* (1280 at ≤96k), that's 26.7/13.3 ms — also not constant.)

Happy to re-encode if the decompile says otherwise on either.

## Verification

```
18854 checks, 0 failures (was 18784 on your tip)
dext target compiles (signing-off check; your signed lane untouched)
```

### DIFF
--- a/ADKVirtualAudioLab/Driver/VirtualAudioDevice.cpp
+++ b/ADKVirtualAudioLab/Driver/VirtualAudioDevice.cpp
@@ -12,6 +12,7 @@
 #include "../Lab/PacketDumpBlob.hpp"
 #include "../Lab/StickyCounterSink.hpp"
 #include "../Lab/VerifyingSlotProvider.hpp"
+#include "../Protocols/Audio/DICE/SaffireIsochLatency.hpp"
 
 using namespace ASFW::Driver;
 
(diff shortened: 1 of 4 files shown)


boggspa:
Closing per review — wrong tree: the lab is the reference; this belongs in the real driver (TimingCursorPolicy is the single-source spot). Redoing against ASFWDriver; the lab-side version stays stashed on boggspa:lab/saffire-latency-model for, well, reference.

PR #27, opened 2026-06-10
audio: lift the Saffire UpdateIsochBufferParams latency model into TimingCursorPolicy

boggspa:
## What

The `UpdateIsochBufferParams` lift, **redone in the right tree** — replaces the misplaced #26 (lab untouched; the lab-side version stays parked on `boggspa:lab/saffire-latency-model` if you ever want it as a rehearsal copy).

## Where it landed

Your `TimingCursorPolicy` already had the trace's numbers hand-planted (`48/128` cursor frames = mode-kLow × 8, `6/16` packets in `Snapshot()`) — so the table became the single source **there**:

- **`Audio/Runtime/SaffireIsochLatency.hpp`** — the trace as a constexpr unit: delay-packet table (`{14,2}/{16,6}/{18,10}/{20,14}`, +2 at 88.2/96k, +4 at 176.4/192k), DMA depth `160/80/40`, frames-per-packet `8/16/32`, derived delay-frame helpers, full provenance (host-internal chain, latencyMode = host preference, DICE-quirk note), unsupported rates rejected.
- **`TimingCursorPolicy`** — `CursorOffsetFrames` and the snapshot's delay packets now derive from the table at `LatencyMode() = kLow`. **Zero behavior change**, proven by `static_assert`s pinning 48/128/6/16. Your bench-tuned ADK values (`ReportedLatencyFrames` 29, `SafetyOffsetFrames` 8) are deliberately untouched — documented as sitting *on top of* the cursor offset, mapping still a bench question.
- **`tests/audio/SaffireIsochLatencyTests`** — every table cell pinned + the policy-derives-from-table single-sourcing check.

## Two prose/table discrepancies in the trace — please adjudicate (code follows the table)

1. *"safest mode = 160 frames"* for the **output** offset at 48 kHz — table says output mode 3 = 14×8 = **112**; 160 = the **input** column (20×8).
2. *"constant ~20 ms hardware buffer regardless of rate"* — 160/80/40 packets × 125 µs = **20/10/5 ms**; only ≤48k is 20 ms.

## Verification

```
host suite: 1217/1217 pass
static_asserts: derived values == previously hand-planted values
```


boggspa:
Closing this — it was built against the pre-DICE-merge tree, and TimingCursorPolicy was rewritten during the audio rewrite, so the integration no longer applies. The Saffire latency table is preserved in git history if useful later. Thanks!

PR #36, opened 2026-07-01
fix(driverkit): add arm64e architecture and fix DV payload trailer length

gdemontalivet:
This PR fixes a critical issue where the DriverKit extension failed to load on Apple Silicon because it was built as standard `arm64` instead of `arm64e` (missing Pointer Authentication ABI).

Additionally, this PR adds support for Video DV streams by fixing the DV payload trailer length calculation in `DVCaptureSink.hpp`, allowing video data from FireWire cameras to be correctly parsed and captured.


ASFW.xcodeproj/project.pbxproj
```diff
@@ -603,6 +603,7 @@
 				CODE_SIGN_STYLE = Automatic;
 				CURRENT_PROJECT_VERSION = 1;
 				DEVELOPMENT_TEAM = F6YA6B56LR;
+				ARCHS = "x86_64 arm64e";
```
mrmidi, line 607: I wound't target x86_64 there - we are on Tahoe or later

mrmidi: this still looks strange to me: my xcoeproject ARCHS haven't ARCH field and builds fine....

ASFW.xcodeproj/project.pbxproj
```diff
@@ -318,6 +318,7 @@
 /* Begin PBXShellScriptBuildPhase section */
 		3A27C5372ECDECB0009CA664 /* ShellScript */ = {
 			isa = PBXShellScriptBuildPhase;
+			alwaysOutOfDate = 1;
```
mrmidi, line 321: Why?

gdemontalivet: That was an accidental empty run script phase left in the project. It has now been completely removed.

ASFW.xcodeproj/project.pbxproj
gdemontalivet: You're right that it usually builds fine! However, we found that even when building directly from Xcode, the architecture can sometimes incorrectly resolve to standard arm64 depending on the selected run destination (e.g., 'Any Mac' vs. a specific physical device) or the Xcode version.

When that edge case happens, Xcode reports a successful build, but the resulting .dext lacks the Pointer Authentication ABI (arm64e) and crashes with an OS_REASON_EXEC / ENOEXEC when macOS tries to load it. Explicitly hardcoding ARCHS = arm64e; just acts as a fail-safe so Xcode can never accidentally compile the wrong ABI for this target. I've removed x86_64 as you requested, so it's strictly arm64e now!

PR #46, opened 2026-07-05
feat(scsi): SCSIControllerDriverKit HBA bridging SBP-2 (generic target, tested on CoolScan 9000)

mhellevang:
## What

Adds the SCSIControllerDriverKit path from the TODO: an `IOUserSCSIParallelInterfaceController` so user-space SCSI clients (SCSITaskLib, e.g. VueScan) can drive an SBP-2 device without porting that device's command set into the driver. First hardware is a Nikon Super CoolScan 9000 ED, driven end to end on Tahoe.

Nothing in the driver is specific to the 9000 — it's just the first device I had to test against. See "Not tied to the 9000" below.

## How

- `SCSIController/` — the HBA (`ASFWSCSIController`), a phantom SBP-2 nub for the provider match, and a process-global bridge (`SBP2BridgeHub` + `SBP2TargetBridge`) that watches discovery, logs into the SBP-2 unit, and routes SCSI tasks to the session.
- Login-state push channel: the SBP-2 side is a separate IOService and can't be reached through the provider chain, so it pushes login up/down to the HBA through a reverse observer on the hub.
- Deferred pre-login INQUIRY. The framework auto-creates and probes target 0 about 3s before the SBP-2 login finishes. Rather than spoof an identity, we hold that probe INQUIRY and replay it with the device's real INQUIRY once login is up.
- SBP-2 command plane, previously login/management only, now handles real command traffic: `ORB_POINTER`-per-command submission (same as Apple's shipping SCSI transport and Linux `sbp2` — neither uses the doorbell for SCSI), dead-bit agent-reset recovery, an ORB-timeout → `AGENT_RESET` → LUN-reset → bus-reset escalation ladder, page-table data descriptors, autosense, and `ISessionScheduler` ORB timers instead of IOSleep-on-queue.
- Teardown: target 0 is auto-created by the framework's presence scan, so `Stop` does not call `UserDestroyTargetForID` on it. Calling destroy re-entered the framework's own termination on the aux queue and hung target quiesce (60s registry busy-timeout panic). Held-INQUIRY flush stays as a backstop.

## Not tied to the 9000

There's no GUID/vendor/model match anywhere. Matching co-matches the OHCI PCI device; SBP-2 unit adoption keys on the standard spec_ID / sw_version; node ID, LUN, speed, ORB size, timeouts and max_payload all come from Config ROM and the bus at runtime. And all the SCSI command generation (INQUIRY, MODE SELECT, LUT upload, READ) is in VueScan, not the dext.

So another single-LUN SBP-2 scanner — a CoolScan 8000, say — should enumerate, log in and run through the same path with no code change. It'd be capable out of the box; only the 9000 is actually validated. The handful of 9000-observed constants and workarounds are commented as such in-code (each is either generic-safe or only fires on a fault).

## Fixed since the first cut

- Identity is no longer hardcoded — real INQUIRY is deferred and replayed post-login (above).
- Task completion is now exactly-once. A synchronous `WriteBlock` failure could complete a task twice (submit-path error plus teardown/abort).
- The bridge holds `SessionRegistry` via `weak_ptr` instead of a raw cross-service reference, closing a use-after-free on teardown (FW-60 class).

## Known limitations (intentional, commented in code)

- `RESERVE`/`RELEASE` are answered synthetically GOOD. This HBA is the only initiator on the bus, so a reservation is uncontended and GOOD is correct for any single-initiator SBP-2 target. (What flagged it: the reference path never sends them, and 9000 firmware also wedges on a `RESERVE(6)` retry.)
- One target at a time — target 0 / LUN 0 / one session. A per-GUID multi-target map is future work; not worth building without multi-device hardware to test it. I'd wager _most_ people only want to run a single device at the time anyway.

## Testing

- Host suite green (1219 tests), including regressions for the exactly-once-completion and cross-service-lifetime fixes. Covers the command plane, page tables, the SBP-2 user-client handler, and the login-state channel.
- End to end on Tahoe: real film scan, transport log clean (0 unclaimed/dropped, dead-bit recovery handled, no wedge). VueScan finds and opens the real LS-9000 via the deferred INQUIRY.
- Teardown, worst case: with a scan running (task in-flight), yanking the Thunderbolt-FireWire link mid-transfer tears down `IOThunderboltPort` and the SCSI target with an outstanding task. `Stop` runs, the in-flight command is force-completed (`kIOReturnAborted`), the framework terminates target 0 — no panic (this exact path used to 60s-busy-timeout-panic before the teardown fix). Reconnecting after a mid-scan yank needs a scanner power-cycle, same as the old macOS FireWire/SCSI stack — that's the 9000 firmware, not us.

## Dependency

Needs the AR-stitching fix (#45), or a full scan wedges at the DMA buffer boundary. #45 is merged, so this is rebased straight onto `main` and no longer stacked.


mhellevang:
Skjermbilde 2026-07-05 kl  11 37 59

Issue #54, opened 2026-07-10
SCSI related kernel panic


mhellevang:
lychzord reports kernel panic due to scsi changes. 

lychzord, can you paste panic report please so I can investigate further? 😄 

1. The panic report (newest .panic in /Library/Logs/DiagnosticReports/) - if IOSCSIParallelFamily is in the backtrace, theory confirmed
2. csrutil status, sw_vers, nvram boot-args
3. Did you re-enable SIP or update macOS after installing?

alicankaralar:
panic(cpu 0 caller 0xfffffe004b7534f4): busy timeout[0], (60s): 'pci11c1,5901' (1e,4020001) @IOService.cpp:5986
Debugger message: panic
Memory ID: 0xff
OS release type: User
OS version: 25F80
Kernel version: Darwin Kernel Version 25.5.0: Mon Apr 27 20:41:26 PDT 2026; root:xnu-12377.121.6~2/RELEASE_ARM64_T8132
Fileset Kernelcache UUID: 810ED0581F8496D5014005FD8C299E58
Kernel UUID: 4FEB3A7D-D020-3910-B7AF-232C78D32651
Boot session UUID: E8453AD4-AE53-4CA0-A3D3-B3D50DC02028
iBoot version: mBoot-18000.120.36
iBoot Stage 2 version: mBoot-18000.120.36
secure boot?: YES
roots installed: 0
Paniclog version: 15
Debug Header address: 0xfffffe00268c9000
Debug Header entry count: 3
TXM load address: 0xfffffe0036824000
TXM UUID: F0C1A57F-CE44-3DF4-9E86-0CFF02D7C00D
Debug Header kernelcache load address: 0xfffffe0046824000
Debug Header kernelcache UUID: 810ED058-1F84-96D5-0140-05FD8C299E58
SPTM load address: 0xfffffe0026824000
SPTM UUID: BF0FEC4B-5AEA-35B8-BF85-518497A9C1C5
KernelCache slide: 0x000000003f820000
KernelCache base:  0xfffffe0046824000
Kernel slide:      0x000000003f828000
Kernel text base:  0xfffffe004682c000
Kernel text exec slide: 0x0000000043f44000
Kernel text exec base:  0xfffffe004af48000
mach_absolute_time: 0x7a5479b3
Epoch Time:        sec       usec
  Boot    : 0x6a50df3f 0x00004284
  Sleep   : 0x00000000 0x00000000
[...]

PR #61, opened 2026-07-11
fix(scsi): create target 0 at SBP-2 login, not at boot — HBA-side fix for #54

mhellevang:
## What

The HBA-side fix for #54: SCSI target 0 is now created at **SBP-2 login** and destroyed at **terminal logout**, instead of being auto-created by the boot-time presence scan. This makes `--scsi` builds cold-boot-safe: a machine booting with no SBP-2 device on the bus has no target whose probe could strand the registry into watchdogd's 60 s busy-timeout panic (`IOService.cpp:5986`).

Builds on the opt-in gate from #55 (merged). Supersedes `main`'s competing held-INQUIRY fix for #54 — rationale in the collapsed rebase note at the bottom.

**Status:** hardware validation is **complete** (2026-07-29 → 07-31, FW adapter + Nikon CoolScan 9000 ED on Tahoe; all 7 scenarios pass, table below). Host suite 1377 green; `--scsi` build verified (personality present in the artifact).

## How to read the commits

Net content is **7 commits**. The other 5 are neutral noise from keeping the branch fast-forward-only (no force-push); happy to squash on merge.

| Commit | What |
|---|---|
| `9aa9d82` | Core model: create target 0 on the login up-edge, destroy on terminal logout; presence scan answers `false` unconditionally |
| `91cd372` | Lifecycle hardening after adversarial review (dedicated serial queue, Stop-deadlock avoidance) |
| `a156372` | Docs: correct the misleading `UserDoesHBAPerformDeviceManagement` comment |
| `6922a4c` | Emit a terminal down-edge when a reconnect is rejected or times out — root cause of data-out 5/26 after device power-cycle |
| `c06ef76` | Pre-destroy any existing target 0 before the login-edge create — fixes the create collision with the family's boot scan |
| `ad44e62` | `TargetReadinessGate`: gate the login-up create on TEST UNIT READY — fixes the childless-target race |
| `19916e3` | Raise the readiness-gate probe budget 90 → 150 |

Skip when reviewing: `e1befce` + `0dfa6df` (ROMScanner fix + its revert — landed separately as #90), `8dd9480` + `e415899` (placeholder removal + its revert — net zero), `1b4b98d` (xcodegen 2.46.0 regeneration — `main` carries the same as `7042779`).

## Model (pure framework hotplug)

- `UserTargetPresentForID` answers **false unconditionally** — the bring-up scan never creates a target, so a login-driven create can never be duplicated by the scan (the HW-observed v49 duplicate-target wedge is dead by construction; anything short of constant false can still be read `true` by a late scan).
- `UserCreateTargetForID(0)` on the login-up edge, `UserDestroyTargetForID(0)` on terminal logout/login-failure — delivered via the existing `SBP2BridgeHub` observer. The registry emits terminal edges only, so a transient bus-reset suspension leaves the target alone (reconnect re-asserts login).
- Lifecycle calls run on a dedicated serial queue: not `AuxiliaryQueue` (the create call is routed through it — calling from it never dispatches) and not the Default queue (it services the framework's target-init upcalls). Stop does **not** wait on this queue — an in-flight create needs the Default queue for its init upcall, so a synchronous Stop would deadlock into the same 60 s panic class; instead a stopping flag gates queued blocks and the queues are released in `free()`.
- The deferred-INQUIRY hold machinery is **removed**: the SAM registration probe runs right after the login-edge create, while the session is up, so it returns the device's real identity. In the suspended window, INQUIRY answers BUSY like everything else — nothing is ever held without a deadline (an indefinite hold is the #54 mechanism relocated: a device that vanishes while suspended emits no terminal edge).

## Hardware validation — all scenarios pass

| # | Scenario | Result |
|---|----------|--------|
| 1 | Cold boot, adapter attached, scanner off | ✅ no panic, no target, `no target until SBP-2 login` |
| 2 | Scanner power-on → login → scan | ✅ `target 0 created` on the login edge; full scan |
| 3 | Scanner off/unplug → destroy | ✅ `target 0 destroyed` observed directly |
| 4 | Restart with adapter attached (original #54 scenario) | ✅ clean boot |
| 5 | Adapter unplug while idle | ✅ no panic |
| 6 | Cold boot with scanner on | ✅ login → create → scan |
| 7 | Bus reset mid-scan | ✅ in-flight READ aborted with a transport error, reconnect succeeds with **no** target edge, next scan clean |

## The three HW findings and their fixes

All found on real hardware 07-29 → 07-31, all fixed and re-verified on hardware. Full chronology and log evidence in the collapsed working notes.

- **`c06ef76` — create collision.** The family's own bring-up scan (`UserDoesHBAPerformDeviceManagement=false`) had already created a target-0 node, so `UserCreateTargetForID(0)` hard-failed `0xe00002bc` on every login edge (apple-oss IOSCSIParallelFamily: `CreateTargetForID` fails when `GetTargetForID(targetID) != NULL`; `DestroyTargetForID` on a missing ID is a no-op). Fix: unconditional pre-destroy before the login-edge create. HW-confirmed: the pre-destroy silently succeeds against the family-created node and the create then succeeds deterministically.
- **`6922a4c` — data-out 5/26 after device power-cycle.** A rejected/timed-out reconnect fell back to a fresh `Login()` silently, so the HBA only saw a new up-edge, kept the stale SAM target alive across the device reboot, and no re-probe happened (the power-cycled scanner had lost its MODE SELECT state — hence 5/26 on every data-out while data-in stayed fine). `LoginSession` now emits a terminal down-edge (`NotifySessionLost`) before the fallback login, matching IOFireWireSBP2's model (failed reconnect = lost session). HW-confirmed: power-cycle produces exactly `login down` → `target 0 destroyed` → `login up` → `target 0 created`, and data-out works on the recreated target.
- **`ad44e62` + `19916e3` — device-readiness race (childless target).** The login-edge create fires ~30 ms after device power-on, while the scanner is still mechanically warming up (TEST UNIT READY → NOT READY 2/04, 2/05); the SAM family publishes the `IOSCSITargetDevice` but never publishes the LUN-0 nub and never re-probes — permanently invisible to SCSITaskLib apps until a re-plug. Fix: a `TargetReadinessGate` in `SBP2TargetBridge` between the registry's login edges and the hub — a fresh up edge polls TEST UNIT READY over the bridge's own task queue (unit attentions → immediate re-probe; warm-up senses 2/04, 2/05 → 1 s pacing; fail-open probe budget) and announces the target only when the device answers ready. Down edges and reconnect re-asserts pass straight through (re-assert probes would consume unit attentions from under a running initiator); operational NOT READY (e.g. 2/3A) announces immediately; fail-open means the worst case reproduces today's behavior, never something worse. 12 new host tests. HW-confirmed: `device ready after 72 probe(s)` through the ~71 s LS-9000 warm-up → LUN-0 nub published, VueScan scans clean without any re-plug (the 72-of-90 margin prompted the budget raise to 150).

## Known accepted gaps

- A lifecycle block that passed the stopping check can still be inside a create/destroy kernel call when the framework begins terminating the controller — the framework must tolerate hotplug calls racing termination (call fails, is logged).
- A device that vanishes while suspended (never reconnects, never reaches a terminal state) leaves an attached target answering BUSY until teardown — cosmetic. Follow-up (not this PR): a suspend timeout (Suspended → Failed after reconnect-hold expiry) so vanished devices reach a terminal edge.
- Aborting the in-flight command at a mid-scan bus reset surfaces as an I/O error to the app (VueScan fails the scan; the next scan is clean). Same semantics as Apple's stack — retry is the initiator's job.

<details>
<summary><b>Rebase note — supersedes main's competing #54 fix (2026-07-2x)</b></summary>

Rebased onto `main` @ `a437b36`. In the interim `main` landed its own #54 fix (`c276546` "Stability improvements" + `2a618db` "improve driver lifecycle diagnostics"): a **held-INQUIRY** approach — target 0 stays present from boot (`UserTargetPresentForID` returns `true` unconditionally), the pre-login probe INQUIRY is *held*, and a new 20 s `holdTimer` + `loginWindowExpired` backpressure bounds the hold so it can't cross the 60 s registry busy-timeout.

This PR takes the **opposite** approach — no target until SBP-2 login (`UserTargetPresentForID` returns `false` pre-login), so the premature probe never happens. The two designs are **mutually exclusive**: they set the same presence bit to opposite values. With presence `false`, main's held-INQUIRY machinery (`holdTimer`, `loginWindowExpired`, `HeldInquiry`) is unreachable dead code, so the rebase **drops it**. Rationale — the login-edge model:

- removes the *cause* (present-but-not-ready target) rather than bounding the symptom's duration;
- kills the v49 duplicate-target teardown wedge by construction (constant `false` can't race the bring-up scan into a duplicate create);
- handles "login never completes" with no target to strand at all.

Main's 20 s timer was a genuine fallback for the post-create suspended window; this PR handles that window explicitly and the target only exists while a login is up, shrinking it. Salvaged from main's commits: the `%{public}s` logging hygiene in `SBP2TargetBridge`. Flagging so the discard of `c276546`/`2a618db`'s SCSI work is a visible, reviewable decision — not a silent merge artifact.

Replaces draft mhellevang#1 (was stacked on the #55 branch in the fork; a PR can't be re-targeted across repos).

</details>

<details>
<summary><b>Working notes — validation chronology, housekeeping (author's tracking, not review material)</b></summary>

## 2026-07-30 — split per review feedback

- The ROMScanner stale-bus crash (found during HW validation, pre-existing on `main`) is fixed in its own PR: #90 (HW-verified and ready). The fix commit is reverted on this branch — HBA and Config ROM work kept separate. Net content = the three login-target commits.
- Also reverted `refactor(scsi): remove unused ASFWSBP2Nub placeholder`: on `dev` the nub is no longer unused (`SBP2NubPublisher` creates one per discovered SBP-2 unit), so deleting it here would only manufacture an integration conflict.
- Reverts instead of a history rewrite to keep the branch fast-forward-only; happy to squash on merge.

## 2026-07-29 session — first HW findings (both since fixed, see above)

Tests 1, 2, 3 (partial), 5, 6 ran on real hardware; scanning worked end-to-end. Findings:

1. **Create collision** — `UserCreateTargetForID(0) failed 0xe00002bc` on every login edge; the family's own bring-up scan (`UserDoesHBAPerformDeviceManagement=false`) had already created a target-0 node. Scanning worked only because the family's INQUIRY retry registered that node after login — a ~1 s window after boot, so the primary #54 path (boot without device, power on later) worked by timing luck. Side effect: `targetAttached` stayed false, so the destroy leg was silently skipped (test 3 caveat). → fixed in `c06ef76`.
2. **Data-out 5/26 after power-cycle re-login** — all data-out (SET WINDOW 0x24, SEND 0x2a, vendor 0xE0) failed 5/26/00 while data-in was fine; log showed "login up" with no new "session created". Counter-proof both ways: cold boot with scanner on scans fine; bus-reset re-login within the same power-on keeps data-out working. → fixed in `6922a4c`.
3. **Dext SIGSEGV on scanner power-on** — pre-existing `main` bug (stale ROMScanner reused over a freed bus after a runtime rebuild), split out and fixed in #90.

## 2026-07-31 session — re-validation (all pass)

- Test 4 (reboot with adapter): no panic, single clean dext instance.
- First-ever successful login-edge create (`target 0 created`, no `0xe00002bc`); clean full scan.
- Pre-destroy breadcrumb logs only on failure — its absence on both login edges means the destroy of the family's boot node **succeeded silently**; shim-vs-family ownership question settled.
- Power-cycle: exact `login down` → `target 0 destroyed` → `login up` → `target 0 created` sequence; scan afterwards clean (three isolated 5/26 at scan start = VueScan renegotiating against the rebooted scanner that lost MODE state, not the persistent 5/26 wall — every affected opcode succeeds on retry).
- Bus reset mid-scan (test 7), triggered in software via the MCP control plane's `asfw_bus_reset_dev` (no second FW device needed): in-flight READ aborted with transport `0xe00002eb`, ~120 ms BUSY window, reconnect success, **no** down-edge, no destroy/create; VueScan failed the in-flight scan (acceptable), next scan completed clean.
- New finding both at cold power-on and at the power-cycle edge: the device-readiness race (childless target) described in the open section above.
- Cable unplug tears down the whole controller (`[SCSIHBA] Stop` → full re-init on replug), not just a logout edge — destroy exercised via both paths.

**8. (optional experiment, non-blocking)** Flip `UserDoesHBAPerformDeviceManagement` to `true` and re-run scenarios 1, 4, and 6. If the bring-up scan disappears from the log (no `UserTargetPresentForID` upcalls), `true` kills the #54 scan at the root — keep the presence guard regardless as defense in depth. If anything regresses, revert to `false`: the current value is the HW-validated configuration.

## Log capture (Tahoe)

Dext lines are live-only on Tahoe (`log show` returns nothing for them):

```bash
log stream --predicate 'eventMessage CONTAINS "[SCSIHBA]" OR eventMessage CONTAINS "[SBP2Bridge]"' --info --debug
```

Keep the predicate narrow — power-on/replug bursts drop lines in `log stream`.

## CI housekeeping

Homebrew's xcodegen is now 2.46.0 and emits the pbxproj group order differently; this branch and #90 carry a one-line regeneration commit. `main` needs the same regeneration on its next change or its CI gate will fail there too.

</details>


mhellevang:
Any thoughts on this @mrmidi ? The pr makes --scsi builds cold-boot-safe

mrmidi:
> Any thoughts on this @mrmidi ? The pr makes --scsi builds cold-boot-safe

Build flags for solving this looks like a band-aid for me, but it will do for now. Doing this properly in code requires too much hassle and refactoring. Otherwise looks like a good changes.

mhellevang:
> > Any thoughts on this @mrmidi ? The pr makes --scsi builds cold-boot-safe
> 
> Build flags for solving this looks like a band-aid for me, but it will do for now. Doing this properly in code requires too much hassle and refactoring. Otherwise looks like a good changes.

Yes, they were a temporary measure. Strictly speaking I don't think they are needed anymore, but I figure we can disable them in a separate PR after I've tested it some more.

PR #68, opened 2026-07-14
Drive ASFW.xcodeproj from XcodeGen

mrmidi:
## Summary

`ASFW.xcodeproj` is now **generated** from a root [`project.yml`](../blob/refactor/xcodegen/project.yml) with [XcodeGen](https://github.com/yonaskolb/XcodeGen) — the project definition becomes reviewable YAML instead of an opaque pbxproj. The generated project stays **committed** (output is deterministic), so CI and plain checkouts build without xcodegen installed; `./build.sh` regenerates it automatically in preflight when xcodegen is present.

- Root `project.yml` claims the standard XcodeGen slot: the stale duplicate ADK spec that lived there (and the committed root `ADKVirtualAudioLab.xcodeproj` it generated) are removed. The real ADK lab spec at `ADKVirtualAudioLab/project.yml` is untouched.
- All the load-bearing project machinery survives verbatim: conditional dext entitlements via `$(ASFW_DEXT_ENTITLEMENTS_$(ASFW_ENABLE_SCSI))`, Info.plist preprocessing for the SCSI personality, `bump.sh` pre-build version stamp, dext embedded unsigned at `Contents/Library/SystemExtensions` with `RemoveHeadersOnCopy` only, schemes + `ASFW.xctestplan`, MCP swift-sdk package pins.

## Deliberate deltas (everything else is 1:1)

- DriverKit frameworks linked via `-framework` flags + SDK search path instead of file refs hardcoding `DriverKit25.0.sdk`/`DriverKit25.1.sdk` paths (same link result; stops breaking on Xcode SDK bumps; same pattern the ADK lab dext already uses)
- `bump.sh` no longer ships as a bundle resource inside the .app and .dext
- Junk the old Xcode-16 synchronized folders silently bundled into the dext is excluded: `fixlog.sh`, `*.md.bak`, `.claude/settings.local.json`
- Vestigial empty run-script phase on the app target dropped; stale `project.pbxproj.backup` and tracked xcuserdata removed

## Verification (A/B against the old project)

- `xcodebuild -showBuildSettings` diff across all 3 targets × Debug/Release: clean — only the framework-flag normalization and project-name noise
- Clean-room builds of old vs new project: **built app and dext Info.plists byte-identical**; bundle file-list diff shows only the approved removals above
- `--scsi` lane: `CODE_SIGN_ENTITLEMENTS` flips to `ASFWDriver+SCSI.entitlements` and the `ASFWSCSIControllerService` personality appears via plist preprocessing
- `./sign.sh` embeds exactly the per-target entitlement sets (verified with `codesign -d --entitlements`)
- Swift XCTest suite passes (proves scheme + test-plan rewiring), C++ CMake/CTest suite passes, `compile_commands.json` covers all 202 dext translation units × 2 arches
- Regeneration is deterministic: `xcodegen generate` on a clean tree produces a byte-identical pbxproj

## Workflow change

Synchronized folders are gone (XcodeGen emits classic groups), so new/renamed source files are **not** picked up automatically: edit `project.yml` (never the pbxproj / Xcode UI settings), run `xcodegen generate` (or just `./build.sh`), and commit the regenerated project alongside. Documented in README and CLAUDE.md.


mrmidi:
Some AV/C Duet fixes applied too.

PR #71, opened 2026-07-17
FW-105: fix AV/C function block CDB encoding, route Phase 88 mixer

mrmidi:
## Summary

- `4ae710b5` FW-105: fix AV/C function block CDB address encoding — `AVC_FUNCTION BLOCK` operands were missing the channel/plug-ID byte that must precede the control selector (AV/C Audio Subunit spec, §7.x FUNCTION BLOCK). Adds `AudioFunctionBlockCommand::BlockType` (Selector 0x80 vs Feature 0x81) and a constructor that takes it explicitly; `AudioSubunit`/`MusicSubunit` `SetAudioVolume` callers and their tests updated to match the corrected wire format.
- `207342e0` FW-105: PHASE 88 BeBoB — configure hardware mixer routing on clock apply — after AV/C unit plugs are set to AM824 48k, chains FUNCTION BLOCK commands (via the new Selector-block support) to route the Phase 88's internal hardware mixer for stereo playback: Mixer Destination (FB 0x06) → analog-output-1/2, Mixer Stream Source (FB 0x07) → stream-input-1/2, then unmutes and sets unity gain on both the stream playback and mixer output feature blocks. Without this the device never routes the incoming AMDTP stream to its outputs.

Builds on the Phase 88 BeBoB backend merged in #70.

## Test plan
- [x] `./build.sh --test-only --test-filter "AudioFunctionBlockCommand|MusicSubunit"` — passes
- [x] Hardware smoke test on Phase 88 — mixer routing (destination/source select + unmute/gain) still needs a HW pass to confirm audio actually reaches the outputs


mrmidi:
Ok, there is something strange with routing I need to research: I've got outputs from out 1 and out 3. I need to dig internal routing for PHASE 88.

PR #72, opened 2026-07-17
FW-64: AV/C RX stream-health recovery — timing-loss debounce + restart

mrmidi:
## Why this exists

On the AV/C backend, a persistent RX replay outage had **no floor**. RX fires a timing-loss
callback when an *established* replay cadence dies (`IsochReceiveContext::ResetReplayEpochForDiscontinuity`
→ `IsochService` → the transport callback). `DiceAudioBackend` subscribes to that callback and
escalates; **`AVCAudioBackend` never did** — it only had bus-reset recovery.

So when replay died mid-session (e.g. an RX gap reset the replay ring), TX legally degraded to
NO-DATA and the content-frame cursor froze while CoreAudio kept writing. The designed transient
exit — the `[TxAlign]` self-heal re-projecting the cursor once replay recovers — works, but if the
outage *persisted*, nothing escalated: the wire stayed NO-DATA/stale and audio stayed dropped until
the device stream spontaneously returned. The motivating incident showed the content-frame frontier
stranded ~1 s behind CoreAudio (`deficitMax≈50712`, `writtenDelta=0`) with a **full** packet ring —
proving it is not a buffer-size problem; no ring size bridges a stranded cursor.

Full doctrine + three-stack (Linux bebob/oxfw, libffado, Apple `AppleFWAudio`) analysis is in
`AVC_STREAM_HEALTH_AND_RECOVERY.md`. This PR implements the §6 design at **AV/C-local scope**
(structured to lift into the shared coordinator later per FW-64).

## How it works

**The health signal for AV/C is the RX packet cadence — not a device register.** All three reference
stacks read cadence/continuity, never device state, to decide stream liveness (doc §5). So:

1. **`IsochService::IsReceiveReplayEstablished()`** forwards the master RX replay state
   (`IsochReceiveContext::IsReplayEstablished()`, already cross-queue-safe — it snapshots `.control`).
   Surfaced through `IIsochDuplexHostTransport` (added as a **non-pure** virtual defaulting to `false`,
   so the coordinator test fake keeps compiling) with a concrete override.

2. **`AVCAudioBackend` subscribes `SetTimingLossCallback(HandleTimingLoss)`** in the ctor; the
   callback is cleared in the dtor and in `BeginTeardown` **before** the queue drain (FW-61 teardown
   discipline). `HandleTimingLoss`:
   - guards on `stopping_`, `activeGuid_ == guid`, and `IsOperationInFlight` (a start/stop/recovery
     already running is itself the transition that tripped the detector — let it settle);
   - dedups via the existing `recoveringGuids_` set, shared with the bus-reset path so a bus-reset
     rebind supersedes;
   - then, on the `com.asfw.audio.avc` queue:
     1. **Settle ~256 ms** (Apple's 80 ms × 2 analog, ≥ several IO windows), polling `stopping_`.
     2. **Verdict = RX cadence:** if `IsReceiveReplayEstablished()` is true, the gap was a host-side
        StartIO/StopIO discontinuity that the `[TxAlign]` self-heal already absorbed → **suppress**
        and reset the escalation budget.
     3. Otherwise → **escalate** via the existing `AudioDuplexCoordinator::RecoverStreaming(guid,
        kRecoverAfterTimingLoss)` — the wire-observable CMP break/re-establish (matches bebob's
        `break_both_connections` + `cmp_connection_establish`, doc §2/§7).
   - **Bounded** by a per-GUID `timingLossAttempts_` budget (max 4, guarded by `lock_`), reset on
     self-heal / restart-success / stop / removal, so a device that only partially returns cannot
     restart-loop forever.

### Why the settle is a *debounce*, not bebob's 1 s warm-up

The 256 ms window is a **pre-restart debounce** — the transient-tolerance every reference stack has
(Apple 2×, FFADO 2 s, Linux jumbo-cycle). It is **not** bebob's ~1 s *post-CMP* NO-DATA warm-up.
Warm-up cannot cause a restart loop here because timing-loss fires only when a **previously
established** replay dies (`wasEstablished` guard); during warm-up replay was never established, so it
never re-fires.

## How this differs from DICE

Same trigger (`SetTimingLossCallback`), same escalation vehicle (`RecoverStreaming`), same
`DiceRestartReason` enum — but a **different health verdict**, because the two device classes expose
different ground truth:

| | DICE (`DiceAudioBackend`) | AV/C (`AVCAudioBackend`, this PR) |
|---|---|---|
| Health probe | **Register read** — DICE GLOBAL/STATUS/EXT_STATUS clock lock + rate (`ReadDuplexHealth`) | **RX cadence** — `IsReceiveReplayEstablished()` after a settle window |
| Why | TCAT exposes a clean lock/rate block | AV/C has **no** such register; the packet stream *is* the signal (doc §5) |
| Debounce | health-gate at event time (locked+healthy ⇒ suppress) | ~256 ms settle, then re-check cadence |
| Escalation | `RecoverStreaming` | `RecoverStreaming` (same) |
| Retry bound | coordinator policy + `recoveringGuids_` | + per-GUID attempt budget (max 4) |

The DICE register probe is a **TCAT-specific luxury and must not be assumed portable** — that is the
whole reason AV/C uses cadence instead. Both paths converge on the same shared coordinator restart.

## Secondary: `[TxPrepRange]` log flood

The `[TxPrepRange]` anomaly line (`ASFWAudioDriverZts.cpp`) flooded at ~1 kHz during the incident
because the persistent `frameShort` stall tripped the gate every wake. It now uses `ASFW_LOG_RL` with
a **runtime** interval of `stoppedShort ? 0 : 1000` ms: `stoppedShort` still logs every wake (rare,
precedes an IT FATAL — kept for diagnosis), while a `frameShort`-only stall throttles to ~1/s with a
suppressed-count. The `[TxPrepRange]` grep tag is preserved.

## Safety / invariants respected

- **No new blocking CMP wait** — only *triggers* the existing coordinator recovery, so FW-94's
  CMP-as-CPS discipline is untouched.
- **FW-61 teardown** — callback cleared before the drain; every queued block re-checks `stopping_`.
- **Cross-queue** — `IsReplayEstablished()` snapshots `.control` (FW-62 pattern); the debounce runs on
  the dedicated `com.asfw.audio.avc` queue.

## Testing

- Xcode dext build: **succeeds** (only pre-existing retain-count analyzer warnings in an untouched
  function).
- Full C++ suite: **1335/1335 pass**, including `AudioDuplexCoordinatorTests.TimingLossRecoveryRestartsRunningSession`.
- **Unverified on hardware** — an actual RX-outage → restart cycle still needs to be observed on a
  device. No unit test for the backend debounce path (needs `IOSleep`/`DispatchAsync`/DriverKit
  mocking; `AVCAudioBackend` has no test file); the coordinator-level recovery is already covered.

### DIFF
--- a/ASFWDriver/Audio/DriverKit/ASFWAudioDriverZts.cpp
+++ b/ASFWDriver/Audio/DriverKit/ASFWAudioDriverZts.cpp
@@ -643,8 +643,12 @@ void IMPL(ASFWAudioDriver, TxPreparationReady)
         if (stoppedShort || frameShort) {
             const uint64_t frameDeficit =
                 frameShort ? (targetFrameEnd - exposedFrameEndAfter) : 0;
-            ASFW_LOG(
-                DirectAudio,
+            // stoppedShort is rare and precedes an IT FATAL -> always log (interval
+            // 0). A frameShort-only wake is a persistent stall (e.g. RX outage
+            // NO-DATA) that otherwise floods at ~1 kHz -> rate-limit to ~1/s, with
+            // the suppressed-count preserving burst visibility. (doc §6)
+            ASFW_LOG_RL(
+                DirectAudio, "tx/prep-range", stoppedShort ? 0u : 1000u, OS_LOG_TYPE_DEFAULT,
                 "[TxPrepRange] retiredAbs=%llu prepareBaseAbs=%llu "
                 "prepareUntilAbs=%llu coverageTargetAbs=%llu limitTargetAbs=%llu reqSpan=%llu "
                 "prepared=%u short=%d firstMissingAbs=%lld marginAfter=%llu "
--- a/ASFWDriver/Audio/Protocols/Backends/AVCAudioBackend.cpp
+++ b/ASFWDriver/Audio/Protocols/Backends/AVCAudioBackend.cpp
@@ -46,9 +46,15 @@ AVCAudioBackend::AVCAudioBackend(AudioNubPublisher& publisher,
                        "AVCAudioBackend: Failed to create recovery queue (0x%x)",
                        queueStatus);
     }
+
+    // Subscribe to RX timing loss. Unlike DICE (which health-gates on a register
+    // probe), AV/C has no health register — HandleTimingLoss reads the RX replay
+    // cadence itself after a debounce. See AVC_STREAM_HEALTH_AND_RECOVERY.md §6.
+    hostTransport_.SetTimingLossCallback([this](uint64_t guid) { HandleTimingLoss(guid); });
 }
 
 AVCAudioBackend::~AVCAudioBackend() noexcept {
+    hostTransport_.SetTimingLossCallback({});
     if (lock_) {
         IOLockFree(lock_);
         lock_ = nullptr;
@@ -91,6 +97,7 @@ void AVCAudioBackend::OnDeviceRemoved(uint64_t guid) noexcept {
         IOLockLock(lock_);
         configByGuid_.erase(guid);
         recoveringGuids_.erase(guid);
+        timingLossAttempts_.erase(guid);
         if (activeGuid_ == guid) {
             activeGuid_ = 0;
         }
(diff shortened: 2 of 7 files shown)


mrmidi:
This stuff requires hardware testing.

PR #73, opened 2026-07-17
ADKLab: add input stream + BeginRead instrumentation for C4

mrmidi:
## What

Adds a duplex **input path** to the `ADKVirtualAudioLab` virtual device (mirrors the existing
output stream/ring) and instruments `BeginRead` so the lab can answer README open question **C4**.

Scope is entirely within `ADKVirtualAudioLab/Driver/VirtualAudioDevice.cpp` — no ASFWDriver / dext
changes.

## Why

The lab previously only exercised `WriteEnd` (output), so it could not observe the capture side at
all. **C4** — *"does the ZTS period / declared safety offset interact with the HAL IO buffer size in
a way that constrains ASFW's choice?"* — was therefore unanswerable except by guessing.

## How

- Adds an input stream + ring mirroring the output path.
- Wires up `SetInputSafetyOffset` / `SetOutputLatency` / `SetInputLatency` (previously only
  `SetOutputSafetyOffset` was ever called).
- Instruments `BeginRead` against a simulated hardware capture cursor advanced independently by the
  ZTS timer.
- Adds a `dump beginread` StopIO log line reporting the **real** coreaudiod-scheduled relationship
  between `BeginRead` and `WriteEnd` host times, instead of a guess.

## Notes

Lab-only, hardware-free (the ADK wind-tunnel). Intended as a testing aid for the ZTS/safety-offset ↔
IO-buffer-size question ahead of the FireWire-side work.

### DIFF
--- a/ADKVirtualAudioLab/Driver/VirtualAudioDevice.cpp
+++ b/ADKVirtualAudioLab/Driver/VirtualAudioDevice.cpp
@@ -47,6 +47,19 @@ namespace {
 constexpr uint32_t kSampleRate = 48000; // the ZTS period math below is 48 kHz-only
 constexpr uint32_t kMaxOutputChannels = 32; // bound for the layout array
 constexpr uint32_t kBytesPerSample = sizeof(float);
+
+// C4 rig: declared per-direction latency/safety-offset under test. Set to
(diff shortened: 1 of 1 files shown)


mrmidi:
Experimental bench update. Not tested yet. For further latency optimizations

PR #77, opened 2026-07-18
docs: de-DICE refactoring plan for the shared audio layer

mhellevang:
## What

A written plan (`DEDICE_REFACTOR.md`) for the boundary-leakage problem: DICE leaking into OXFW/BeBoB paths through the shared audio layer. No code changes — plan only, so we can agree on the approach before renames land.

## Baseline update

The first version was written against `main`; it turns out a good chunk of the cleanup already exists on branches that haven't landed (no open PRs against `main`, so not visible from a plain pull). Rebased the plan on that work:

- **`codex/fw-73-neutral-duplex-rename`** (incl. PR #67 / FW-73b) already neutralizes the restart/session vocabulary (`DuplexRestartReason` & co. in `Duplex/DuplexRestartSession.hpp`, alias veneer gone, `AVCAudioBackend` clean). Note: #67 merged into the fw-73 branch, not `main` — the branch itself hasn't landed.
- **`refactor/BeBoB`** implements the `BeBoB_REFACTOR.md` plan.
- The two branches don't include each other, and fw-73 is based on `main` as of 2026-07-11, so landing both means resolving mechanical conflicts in `AudioDuplexCoordinator`/`AVCAudioBackend`.

**Phase 0 of the plan is now: land those two branches.** Everything else is scoped to what remains after that.

## What remains (the actual plan)

1. **Finish recovery/restart cleanup** — raw DICE register words (`notification`/`status`/`extStatus`) on the neutral result structs become a protocol-opaque diag struct; `DiceRecoveryPolicy` → `StreamRecoveryPolicy` (`RecoveryPolicyReason`/`RecoveryDisposition`).
2. **TX engine** — `DiceTxStreamEngine` → `AmdtpTxStreamEngine`, neutral namespace (it's a generic AMDTP engine in all but name).
3. **Neutral recovery dispatch** — new `BackendRecoveryEvent` + `IAudioBackend::HandleRecoveryEvent`; the DICE branches in `AudioCoordinator` move into `DiceAudioBackend`. Scoped to recovery only, not full lifecycle.
4. **Profile contract** — hoist the neutral part of `IDiceDeviceProfile` (Apogee/BeBoB profiles stop inheriting it); `DefaultAmdtpProfile` fallback with byte-identical geometry to `GenericDiceProfile`.
5. **Family homes + enforcement** — `DiceAudioBackend` moves into the DICE family; CI boundary check with an explicit composition-root allowlist (factory/coordinator/registry may know all families; shared machinery may not).

Three recovery types fall out: `DuplexRestartReason` (session cause — done on fw-73b), `RecoveryPolicyReason` (policy explanation), `BackendRecoveryEvent` (external events).

Zero wire-observable change intended; only the recovery-dispatch phase carries real regression risk and gets the HW smoke test.

Happy to adjust phases/naming — that's what this PR is for.

### DIFF
--- a/DEDICE_REFACTOR.md
+++ b/DEDICE_REFACTOR.md
@@ -1,0 +1,263 @@
+# De-DICE Refactoring Plan
+
+## Problem
+
+The Apogee Duet (OXFW) came first, but it was simple enough not to need heavy
+shared machinery. The DICE buildout is what created the shared duplex/restart/
+recovery machinery and the AMDTP TX engine — so the "generic" audio layer grew
+up inside DICE names and namespaces. The seams themselves (`IDeviceProtocol`,
+`IAudioBackend`, `IAudioDeviceProfile`, `AudioDuplexCoordinator`) are fine —
+BeBoB and OXFW do plug into them. What was never done is de-DICE-ing the shared
+surface, so every new family transitively inherits DICE plumbing.
+
+`BeBoB_REFACTOR.md` explicitly deferred this ("Full `DuplexControlTypes.hpp`
+cleanup" under *Future Flags*). This plan is that remaining work.
+
+### Leak inventory
+
+| # | Site | Leak |
+|---|------|------|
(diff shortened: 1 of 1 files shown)


mrmidi:
some nice catches here on unfinished work.

PR #79, opened 2026-07-19
Research: stream stability geometry and recovery telemetry

mrmidi:
## Research scope

Captures the Duet/Focusrite stream-stability investigation so contributors can review the cursor, replay, and recovery assumptions together.

## What changed

- Adds CoreAudio write-frontier → TX-preparation signalling and telemetry.
- Adds compact MCP-ring fault records, including the precise `RxSequenceReplayReader::TryRead()` rejection reason.
- Adds a TX data-horizon burst simulator and updates buffer-geometry verification.
- Documents the observed cursor invariants, replay-horizon failure, FW-64 callback overwrite, and unsafe recovery escalation.

## Current conclusion

This is intentionally **research, not merge-ready**. The 400-cycle TX preparation experiment survives scheduler bursts in simulation, but it incorrectly consumes a finite historical RX replay ring as future TX timing. That can produce NODATA, `TxAlign` jumps, clicks, and silence even while DMA/SYT remain alive.

The PR also identifies two recovery defects introduced by FW-64:

- AV/C overwrites DICE's single timing-loss callback.
- AV/C may restart transport without quiescing/re-priming CoreAudio's TX producer under the same epoch.

## Validation

- `./build.sh --no-bump` (Debug build)
- `python3 tools/tx_data_horizon_burst_sim.py suite`
- `python3 tools/buffer_geometry_verify.py verify --max-io 1024`
- `git diff --check`

The local Xcode scheme rewrite is intentionally excluded.

### DIFF
--- a/ASFWDriver/Audio/DriverKit/ASFWAudioDriverIO.cpp
+++ b/ASFWDriver/Audio/DriverKit/ASFWAudioDriverIO.cpp
@@ -186,6 +186,32 @@ kern_return_t InstallIOOperationHandler(IOUserAudioDevice& audioDevice,
                 control->client.PublishWriteEnd(sampleTime, hostTime, ioBufferFrameSize);
                 PublishPlaybackRingWriteEnd(driverIvars->runtime.directAudioGraph, *control);
 
+                // Keep packet preparation driven by the CoreAudio write
+                // frontier as well as the OHCI refill path. The target is a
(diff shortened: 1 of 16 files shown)


mrmidi:
NOTE, I'm merging it, but currently streaming stability is broken for all devices/backends. If I couldn't resolve it - I will roll back to some working commit.

PR #83, opened 2026-07-20
RME Fireface 800: Stage 5C-5H bounded circular playback integration

Clip4food:
## Summary

- Adds a full RME Fireface 800 device protocol (`ASFWDriver/Audio/Protocols/RME/`) and the no-CIP (tag=0) isoch wire-format support it depends on across `Audio/Wire/AMDTP`, `Shared/Isoch/IsochAudioTransport.hpp`, and the Isoch transmit path (`IsochTxDmaRing`, `IsochTransmitContext`, `IsochService`).
- Discovery: `FWDevice`/`DeviceRegistry` recover `model_id` from the Unit Directory (RME publishes it there, not the root directory).
- Core Audio: publishes a read-only "RME Fireface 800 (Stage 5C)" endpoint (192 kHz, 12 in/12 out); `StartIO` deliberately stays rejected (`kIOReturnUnsupported`) after a single bounded Stage 5H integration burst — no live streaming yet.
- Stage 5H: reuses the Stage 5G-validated 48-descriptor D-D-D-S silence cadence as a circular ring, runs it for a bounded 100 ms with RUN/no-interrupt polling, then tears down OHCI → FF800 → IRM in order.
- Fixes a real bug found during hardware testing: `ReservePlaybackResourcesPreflight`'s IRM snapshot read raced `IRMClient`'s own IRM-node resolution right after a bus reset (`NotFound` while `irmNodeId_ == 0xFF`), permanently stranding the playback route with no retry. Now retries up to 10x/50ms, mirroring the existing `PollDeviceTxIsochChannel` idiom.

## Heads up for review

- Branched off `405cb99`; `main` has since moved ~18 commits (notably `refactor(isoch): make transmit queue payload opaque` and the isoch receive-seam refactors), so this will likely need a rebase/conflict pass before merge — flagging so it's not a surprise.
- I'm not the repo owner and don't have write access, so this comes from a fork.

## Test plan

- [x] `./build.sh --test-only` — 1334/1335 C++ tests pass (1 pre-existing unrelated failure: `IsochDMAMemoryManagerTest.PayloadSlicingAndPageAlignment`, page-alignment assumption in the host-test memory stub).
- [x] `./build.sh --config Release` (Xcode, x86_64) + `./sign.sh` — builds and signs cleanly.
- [x] Installed on real hardware (TI 104C:823F + RME Fireface 800, GUID `0x000a3500f021a1b4`): Stage 5H's bounded circular silent playback burst (48 descriptors, 36 data packets, 12 skips, 100 ms) passed cleanly across a cold boot, a protocol reinit, and a forced bus reset — clean OHCI/FF800/IRM teardown each time, zero event errors, zero DEAD.
- [ ] The specific IRM-`NotFound`-retry branch fixed above did not get re-exercised live (the race window is narrow and didn't reproduce on demand) — verified by code inspection and the original incident log instead.


mrmidi:
Hi and thanks for contribution! I'm looking into it. I want to finish some architecture shifts before actually merging/resolving conflicts and fixing some critical bugs in FireWire driver itself, so please wait some time.

Dreambrother7:
Heads-up on a collision: I've just landed MOTU protocol-v2 support on a fork branch and
we've independently solved the same problem — device identity that isn't in the root
directory — with two different mechanisms, in the same function.

Your side: `FindUnitModelIdInRawROM` + the `PopulateDeviceIdentity` change, recovering
`model_id` from the unit directory because RME publishes it there.

My side ([`1c6ab94`](https://github.com/Dreambrother7/ASFireWire/commit/1c6ab94)): I added
`unitSpecId`/`unitSwVersion` to
[`DeviceProfileQuery`](https://github.com/Dreambrother7/ASFireWire/blob/feature/motu-wire-codec/ASFWDriver/DeviceProfiles/Common/DeviceProfileTypes.hpp#L36-L37)
and pass them through at
[`DeviceRegistry.cpp:52-53`](https://github.com/Dreambrother7/ASFireWire/blob/feature/motu-wire-codec/ASFWDriver/Discovery/DeviceRegistry.cpp#L52-L53),
so the profile providers can match on the unit directory directly
([MOTU provider](https://github.com/Dreambrother7/ASFireWire/blob/feature/motu-wire-codec/ASFWDriver/DeviceProfiles/Audio/Vendors/MotuAudioProfiles.hpp)).

Both touch `PopulateDeviceIdentity`/`MaybeInferKnownIdentityFromGuid` in
`Discovery/DeviceRegistry.cpp`, so whoever rebases second gets a conflict — flagging it now
rather than at merge time.

Worth noting the two aren't equivalent, and I don't think either subsumes the other:

- Your model_id recovery would in fact work for MOTU — the 828mkII does publish
  `Model_Id 0x101800` in its unit directory (I captured its ROM off hardware this week).
- But MOTU's *authoritative* discriminator isn't model_id. Linux matches the whole family
  on `VENDOR_ID | SPECIFIER_ID | VERSION` — see `SND_MOTU_DEV_ENTRY` / `motu_id_table` in
  [`sound/firewire/motu/motu.c`](https://github.com/torvalds/linux/blob/master/sound/firewire/motu/motu.c)
  (motu.c:151-181). `Unit_Sw_Version` is what separates 828mkII (`0x000003`) from Traveler
  (`0x000009`) etc.; the unit model_id isn't used for matching there at all.

So the general shape is probably "surface the whole unit directory identity to the profile
layer, and let each provider pick the field its family actually keys on" — model_id for
RME, sw_version for MOTU. If that sounds right I'm happy to rebase mine onto yours (or
the other way round, whichever lands first) and fold both into one mechanism.

@mrmidi — no action needed from you unless you want to call the shape; just didn't want two
contributors quietly diverging on the same discovery seam.

mrmidi:
> Heads-up on a collision: I've just landed MOTU protocol-v2 support on a fork branch and we've independently solved the same problem — device identity that isn't in the root directory — with two different mechanisms, in the same function.
> 
> Both touch `PopulateDeviceIdentity`/`MaybeInferKnownIdentityFromGuid` in  
> * Your model_id recovery would in fact work for MOTU — the 828mkII does publish
>   `Model_Id 0x101800` in its unit directory (I captured its ROM off hardware this week).
> * But MOTU's _authoritative_ discriminator isn't model_id. Linux matches the whole family
>   on `VENDOR_ID | SPECIFIER_ID | VERSION` — see `SND_MOTU_DEV_ENTRY` / `motu_id_table` in
>   [`sound/firewire/motu/motu.c`](https://github.com/torvalds/linux/blob/master/sound/firewire/motu/motu.c)
>   (motu.c:151-181). `Unit_Sw_Version` is what separates 828mkII (`0x000003`) from Traveler
>   (`0x000009`) etc.; the unit model_id isn't used for matching there at all.

Great findings. Basically that's why Discord server exists for — to share the knowledge and solve the same problems together. :)

PR #92, opened 2026-07-31
ci: build the SCSI-enabled variant on every PR

mhellevang:
## What

Adds a parallel CI job that builds the driver with `ASFW_ENABLE_SCSI=YES` (unsigned) and asserts the `ASFWSCSIControllerService` personality actually landed in the built dext's `Info.plist`.

## Why

The SCSI HBA is opt-in (`./build.sh --scsi`, #55) and OFF in the default CI build, so the two things the flag controls — the `Info.plist` personality preprocessing and the SCSI entitlement selection — are never exercised by CI today. A regression there only surfaces on an opt-in user's machine, where the failure mode is the #54 boot panic. With this job the gated configuration can't silently rot while the flag stays default-off (and if SCSI ever becomes default-on, the coverage is already in place).

The personality assertion uses `PlistBuddy` against the built dext, so a silent preprocessor regression (personality quietly missing) fails CI rather than passing as a "successful" build.

## Housekeeping

First commit regenerates `ASFW.xcodeproj` with xcodegen 2.46.0 (Homebrew's current version, which changed the emitted group order). Without it the existing "Verify ASFW.xcodeproj matches project.yml" gate fails for **every** PR against main — #61 and #90 already carry the same one-line regen on their branches.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -163,3 +163,50 @@ jobs:
       # - name: Build Driver
       #   run: |
       #     ./build.sh
+
+  # The SCSI HBA is opt-in (ASFW_ENABLE_SCSI, ./build.sh --scsi) and OFF in the
+  # default build above, so the Info.plist personality preprocessing and the
+  # SCSI entitlement selection are never exercised by the main job. Build the
+  # variant here so the gated configuration cannot rot while the flag stays
+  # default-off — a malformed personality only surfaces on an opt-in user's
+  # machine otherwise, and the failure mode there is a boot panic (#54).
+  build-scsi:
+    name: Build (SCSI HBA enabled)
+    runs-on: macos-26
+    steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -163,3 +163,50 @@ jobs:
...
+            build \
+            ASFW_ENABLE_SCSI=YES \
+            CODE_SIGNING_ALLOWED=NO CODE_SIGN_IDENTITY= CODE_SIGNING_REQUIRED=NO \
+            > xcodebuild-scsi.log 2>&1 || {
+              echo "::error::SCSI-enabled build failed! Showing last 100 lines of log:"
+              tail -n 100 xcodebuild-scsi.log
+              exit 1
+            }
+
+      # The personality is injected by Info.plist preprocessing; assert it
```
mrmidi, line 196: Looks too safe for me, but let's leave it as is :)

PR #93, opened 2026-08-02
feat(dv): restore capture and harden DriverKit lifecycle

gdemontalivet:
## Summary

- restore DV capture through the upstream payload-neutral `IIsochReceiveConsumer` seam, keeping isoch transport payload-opaque
- negotiate CMP/PCR connections with validated bus-order encoding and broadcast fallback, while preserving upstream's per-device lease ownership
- stop DV capture on bus reset and quiesce receive consumers before buffers, mappings, or hardware are detached
- assemble complete PAL/NTSC DIF frames in the app and write them off the capture hot path
- expose capture discontinuity/drop diagnostics without forwarding corrupt or incomplete frames
- verify DriverKit activation/deactivation against the I/O Registry and detect orphaned servers
- surface live UserClient connection state in Overview and require monotonic installable build numbers
- restore explicit `x86_64`/`arm64e` DriverKit architectures and disable hardened runtime for local ad-hoc dext builds
- merge current upstream `main`, regenerate the Xcode project from `project.yml`, and keep the DV work compatible with the newer multi-stream/CMP architecture

## Validation

- `./build.sh --no-bump`: Xcode build passed
- full C++/CTest suite: 1,385 tests passed (6 reference-fixture tests skipped)
- `./build.sh --swift-test-only`: full Swift/XCTest suite passed, including DV frame assembly and DriverKit health policy tests
- targeted CMP/PCR/DV suite: 18 tests passed
- receive timestamp decoding cross-checked against the local Linux OHCI and Apple IOFireWireFamily references
- hardware-tested on macOS Tahoe with build 3
- recorded PAL DV successfully from a camcorder
- verified clean deactivation/reinstallation and live UserClient status
- verified camcorder disconnect/reconnect discovery

## Local artifacts

The captured `test.dv` file and local build/log scratch files are intentionally not included.


mrmidi:
Hi, thanks a lot for contributing. I'm still looking to your PR but my special thanks goes for taking swift concurrency/static analyzer warnings: I never had a time to fix it working on other problems :)

PR #94, opened 2026-08-04
scsi: match HBA personality on ASFWSBP2Nub — makes the #54 boot panic impossible by construction

mhellevang:
## What

Three commits that remove the *reason* the SCSI HBA is build-gated behind `--scsi`:

1. **`0d700f4`** — the HBA personality matches the already-published `ASFWSBP2Nub` instead of `IOPCIDevice`. A failed dext launch (e.g. AMFI) becomes a silent no-load instead of the #54 boot panic — the panic is impossible by construction.
2. **`7258268`** — publish `iommu-parent` (a DART mapper id) on the nub. Without it the kernel SCSI shim has no DMA mapper and fails every transfer larger than one contiguous page before the dext ever sees the task. The value ships in Info.plist and is machine-specific — see the open question.
3. **`d8fd558`** — emit the logout edge from `SBP2TargetBridge::Shutdown()` so target 0 is destroyed before system sleep. Fixes an unrecoverable post-wake LUN wedge; the wake login rebuilds the target fresh.

## HW validation

Full matrix green on M1 Pro / Tahoe / Nikon LS-9000 via TB→FW dock: large transfers (3× 519 KB READs, VueScan image), the #54 window, sleep/wake ×2, unplug + teardown under load, power-cycle re-login, and a port-swap.

<details><summary>Full matrix</summary>

- Match/launch on nub: `DK: using existing server … from provider ASFWSBP2Nub` — full HBA start, readiness gate, target create with real INQUIRY identity
- Large transfers: 3× 519 640 B READs full payload, 4× 32 KB SEND — VueScan preview produced an image (fails 100 % without the mapper property)
- Reboot with adapter present, scanner off: no panic, clean HBA start against the nub (the #54 window)
- Sleep/wake ×2: `target 0 destroyed` at quiesce → re-login → `target 0 created` after wake; no kernel-side accumulation (1 target, 1 user client after both cycles)
- Adapter unplug under worst case (wedged target + stale client + queued task, pre-fix build): clean cascade teardown ~1 ms, clean rebuild on re-plug
- Teardown under load: in-flight 508 944 B READ aborted cleanly, BUSY on retries, destroy → re-login → re-create → next scan ran
- Power-cycle re-login: no 5/26 regression (the #61 root-cause fix holds on this build)

**Known behavior:** a SCSITaskLib client running across a sleep (VueScan) hangs — its user client points at the destroyed target — and must be restarted after wake; the restarted app finds the device immediately. This is the deterministic, recoverable replacement for the previous unrecoverable LUN wedge.

**Port-swap result: the published DART mapper id is port-independent.** This settles the main open question for the plist path.

Test (MacBook Pro 16" M1, J316, 3 TB ports):

1. Reinstalled the plist variant (`iommu-parent = <57000000>` = `mapper-apciec0-piodma`, baked into `ASFWSBP2NubProperties`) + reboot. Nub published at boot with the property from birth — no app involvement. Baseline VueScan scan on the original port (card under `apciec0`): ✅
2. Moved the adapter to a different TB port, plist untouched. Card re-enumerated under **`apciec1`** (verified in ioreg), nub/HBA rebuilt cleanly, nub still carries the *apciec0* mapper id. Full VueScan preview incl. the ~519 KB READs: ✅

So the shim only needs *a* valid published system mapper to satisfy its DMA-prep path; the actual DMA runs through the PCI device's own mapper regardless of which id the nub advertises. The value is not port-bound.

Consequence for the portable-provisioning question: the "app patches the plist + re-sign + re-install" path (a) is now a **one-time per-machine calibration** at install time (which requires a reboot anyway) — moving the adapter between ports afterwards costs nothing. That removes the reboot-per-port objection. Runtime provisioning into the birth dict remains impossible on current xnu (see the `originalProperties` snapshot analysis above), so (a) looks like the pragmatic default while the DTS question / split-dext design is pending.
</details>

## The open question (why draft)

The mapper id only works from the nub's **birth** properties, and xnu snapshots those from the signed Info.plist when the driver attaches — so runtime provisioning is impossible (HW-falsified twice, root cause in xnu source). The plist value is port-independent (HW-proven), but machine-specific ⇒ per-machine re-signing at install ⇒ incompatible with Developer ID distribution.

<details><summary>Root cause + falsification evidence (xnu source refs)</summary>

**Update: app-assisted runtime provisioning is falsified — root cause found in the xnu source. The mapper id must be present in the dext's Info.plist; no runtime mechanism can deliver it.**

## What was tested (HW, reboot-verified)

Two variants of delivering the app-resolved mapper id to the SBP-2 nub, both on the deferred-publish build (nub only created after the app provisions via the new user-client selector):

1. **Adopt-in-Start:** nub's `Start()` copies the staged id onto itself via `SetProperties` before `RegisterService()`. Property verifiably lands top-level on the nub (`ioreg -d1`), synchronously, *before* the kernel stub is even matched. → **Large I/O still fails** (32 KB SEND / 519 KB READ → instant failure, nothing reaches the dext; threshold between 91 B and 32 KB = the no-mapper single-segment signature). Kernel log completely silent — no DART/prepare errors, i.e. the shim's silent no-mapper cold path.
2. **Bake-before-Create:** `ProvisionSBP2MapperID` patches `iommu-parent` into the `ASFWSBP2NubProperties` dictionary on the driver's own registry entry (verified present in ioreg) *before* calling `Create()`. → **Identical failure.**

Meanwhile the plist variant (value in `ASFWSBP2NubProperties` in Info.plist) remains the only configuration that has ever passed large I/O on the nub provider (519 640 B READs, image in VueScan — verified twice across reboots).

## Root cause (xnu, `iokit/Kernel/IOUserServer.cpp` @ main)

1. **`Create()` never reads the live property table.** `IOService::Create_Impl` fetches the properties dict from `reserved->uvars->originalProperties` — a snapshot taken in `IOUserServer::serviceAttach` (`dictionaryWithProperties()`) when the *driver* service attached at boot. A later `SetProperties` updates the live table (which is what ioreg shows) but never that snapshot. So a runtime-computed property can never enter a `Create()` birth dict — the snapshot chain always bottoms out in Info.plist content, at any nesting depth.
2. **The SCSI shim does not do a live `copyProperty` either.** In variant 1 the property was top-level on the nub before the stub matched (SetProperties_Impl applies synchronously inside the RPC; ring log proves the ordering), and the shim still ran mapperless. The consistent explanation is that the shim consumes the nub's *birth state* (its own `originalProperties` snapshot, taken during `serviceAttach` inside `Create_Impl`, i.e. before the nub's dext-side `Start()` ever runs). This matches all four data points: plist-at-birth works (×2), adopt-in-Start fails, bake-before-Create fails.

`IOMapper::copyMapperForDeviceWithIndex` itself (`iokit/Kernel/IOMapper.cpp`) is a plain `copyProperty("iommu-parent")` + `waitForMatchingService(IOMapperID)` — the caching/snapshotting is on the caller's side.

## Where this leaves the open question from the PR body

- The **mechanism** stands: `iommu-parent` (OSData `IOMapperID` of any published DART) on the nub satisfies the shim's DMA prep — but only when present in the nub's **birth dict**, which on current xnu means **present in Info.plist when the driver attaches**.
- **Portable delivery options:**
  - **(a) App-patches-the-plist:** the app already resolves the correct id portably (class `IODARTMapper`, named by its `IODARTMapperNub` parent `mapper-apciecN-piodma`; tiered fallback; committed with tests). Install flow: resolve → write into the dext copy's Info.plist → re-sign → install → reboot. Uses only the HW-proven mechanism; cost is one reboot when the adapter moves to a different Thunderbolt port. My candidate for the ad-hoc distribution tier.
  - **(b) DTS:** the question sharpens considerably with source line numbers — "what is the sanctioned way for an `IOUserService`-provided `IOUserSCSIParallelInterfaceController` to obtain a DMA mapper on Apple Silicon, given `Create_Impl`'s `originalProperties` snapshot?"
  - **(c)** (speculative, unverified) runtime `IOCatalogueSendData` of a patched driver personality + re-match — avoids the reboot but is AMFI-exposed.

The provisioning selector and deferred publish stay useful for (a)'s UX (the app knows whether the installed plist id matches the current port and can prompt for reinstall), but I've dropped the ineffective bake from the branch. The app-side resolution fix is committed.
</details>

**Direction needed — your call:**

1. **Split-dext** — does your HBA/ConfigROM split address this? Any dext-published nub hits the same birth-properties wall; it only helps if the HBA dext matches something with real PCI ancestry, which reopens the #54 gating question.
2. **DTS incident** — ask for the sanctioned way for an `IOUserService`-provided HBA to get a DMA mapper on Apple Silicon. We can draft it.
3. **Stay on PCI-match** (main's model, mapper for free) and keep #54 solved by gating alone.

Secondary, once direction is settled: default-on / removing the `--scsi` gate (no new regression path — under AMFI the whole dext dies silently either way), and Intel (VT-d publishes a system mapper; property likely unnecessary, untested).

### DIFF
--- a/ASFW.xcodeproj/project.pbxproj
+++ b/ASFW.xcodeproj/project.pbxproj
@@ -226,7 +226,9 @@
 		9939EA3C38D005D2B1706109 /* LockCommand.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 647BC02DDA3AE383EBFED56E /* LockCommand.cpp */; };
 		9A00ED30A11DABC7DD49E7BC /* DriverConnector+Isoch.swift in Sources */ = {isa = PBXBuildFile; fileRef = 88363EE0072BC820ECBB6C1D /* DriverConnector+Isoch.swift */; };
 		9B8E94EA1F0A961CD016D197 /* DriverViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7892E16293D409FFB4F00710 /* DriverViewModel.swift */; };
+		9BD95333AF99D4D79CD5A3DF /* DriverConnector+SCSIMapper.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4DEF4DDB2D9566CAEAFFAAB7 /* DriverConnector+SCSIMapper.swift */; };
 		9BFABEBFD4EE64F02ADFBBF1 /* AsyncSubsystemInterrupts.cpp in Sources */ = {isa = PBXBuildFile; fileRef = B3F76378A96EF757865E11C4 /* AsyncSubsystemInterrupts.cpp */; };
+		9C590D10F760AFC85C707998 /* DARTCandidateSelectionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = FA0CE0102B0AEA2853C383FA /* DARTCandidateSelectionTests.swift */; };
 		9C87DF69A89E62030306D1FB /* DiagnosticsTextFormatter+Topology.swift in Sources */ = {isa = PBXBuildFile; fileRef = 76F0E726223068E61ED92CE8 /* DiagnosticsTextFormatter+Topology.swift */; };
 		9EDB35B4B0F9F503A82DB7A8 /* AddressSpaceManager.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 27D4308D539CCEC3A5EBCFF7 /* AddressSpaceManager.cpp */; };
 		A040141317C828BDF58DB6A6 /* IsochHandler.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 0AC0E4BEE81FF4834F182189 /* IsochHandler.cpp */; };
@@ -641,6 +643,7 @@
 		4D671C398E07E582F14F5A56 /* ControllerCoreAccess.hpp */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.cpp.h; path = ControllerCoreAccess.hpp; sourceTree = "<group>"; };
 		4D994BDB948D370E6F61467F /* DuetControlView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = DuetControlView.swift; sourceTree = "<group>"; };
 		4DB5C96A84E71BA9F2E5835A /* PreSonusStudioLiveProfile.cpp */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.cpp.cpp; path = PreSonusStudioLiveProfile.cpp; sourceTree = "<group>"; };
+		4DEF4DDB2D9566CAEAFFAAB7 /* DriverConnector+SCSIMapper.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "DriverConnector+SCSIMapper.swift"; sourceTree = "<group>"; };
 		4DFC5D271E3484233BD1B0D2 /* TopologyHandler.hpp */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.cpp.h; path = TopologyHandler.hpp; sourceTree = "<group>"; };
 		4E065C1061BFE1DDEBFCEF3B /* ControllerStateMachine.hpp */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.cpp.h; path = ControllerStateMachine.hpp; sourceTree = "<group>"; };
 		4EAC9531EBCD3BFCC5274B38 /* ATContextBase.hpp */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.cpp.h; path = ATContextBase.hpp; sourceTree = "<group>"; };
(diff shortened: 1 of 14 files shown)


mhellevang:
### Note on `dev`'s `SBP2NubPublisher`

Took a look at `dev` to make sure this PR doesn't collide with work already done there — it doesn't, and the per-unit nub direction actually strengthens the approach here:

- **The personality change composes cleanly.** `SBP2NubPublisher`'s per-unit nubs carry the same `ASFWNubType=SBP2` property, so the `IOPropertyMatch` personality in this PR matches them as-is. Per-unit is strictly better together with it: the nub only exists while a device does, so the HBA is only instantiated when needed — which also resolves the "always-published nub blocks in-place dext upgrade" caveat I hit during HW validation.
- **The `iommu-parent` mechanism ports directly:** same property, just set in `SBP2NubPublisher::EnsureNub` instead of at `StartRuntime`.
- **Heads-up on future reconciliation:** `dev`'s SCSI layer predates the #55/#61 merges (held-INQUIRY model, `TargetReadinessGate` removed), so whenever `dev` and `main` meet, `ASFWSCSIController`/`SBP2TargetBridge` will need reconciling — and this PR's sleep fix sits on main's login-driven model. Happy to help with that merge when the time comes.

mhellevang:
Closing this in favor of #96. Summary of why the nub-match direction is dead:

1. **The mapper-id delivery problem is unsolvable under the distribution model.** The `iommu-parent` value only works from the nub's birth properties, birth properties can only come from the signed Info.plist (xnu `originalProperties` snapshot — falsified twice on HW, root cause in `IOUserServer.cpp`), and the value is machine-specific. Central signing + per-machine plist values are mutually exclusive; asking users to re-sign locally is off the table.
2. **The premise itself didn't survive Q1.** Deep-dive into the #54 mechanism (kernelcache disassembly + xnu source, full writeup in #96): the panic is xnu's generic 60 s registry-busy watchdog reached by *stalling* in the shim's synchronous start chain — not a dext-launch-failure panic, and the family kext has no panic sites at all. Nub-match avoided a symptom; a bounded start chain removes the cause. Main's login-driven-target model already has the bounded property, now HW-proven in the exact #54 window (cold boot, adapter in, no device → `busy 0 (1 ms)`, no panic).
3. **The PCI provider supplies the DART mapper for free** — with PCI-match the entire mapper mechanism this PR built becomes unnecessary (large READs verified with no `iommu-parent` anywhere).

The one piece worth keeping — the sleep-wedge fix (`d8fd558`) — is cherry-picked into #96 and re-verified there. The mapper findings (birth-properties snapshot, port-independence, the app-side DART resolution) are preserved in this PR's history if the split-dext design ever needs them.

PR #96, opened 2026-08-05
scsi: ship the SCSI HBA in every build — #54 mechanism corrected, boot window HW-validated (supersedes #94)

mhellevang:
This PR makes the SCSI HBA part of every build by removing the `--scsi` gate, based on two things: the actual #54 panic mechanism is now understood, and the boot scenario it describes has been validated on hardware. It also fixes a sleep/wake bug found during that validation. (Supersedes #94 — the closing comment there explains why the nub-match direction is dead; the PCI provider turns out to supply the DMA mapper for free.)

## What changes

- **The SCSI personality and `scsicontroller` entitlement ship in every build** (`c853e78`, `eefa9f5`). The `--scsi` flag, the Info.plist preprocessing, and the dual entitlements files are deleted; CI now asserts the personality is *present* in the built dext; the README section is rewritten as plain usage documentation.
- **Sleep/wake fix** (`3adfecb`): `SBP2TargetBridge::Shutdown()` now emits the terminal logout edge, so target 0 is destroyed before system sleep and rebuilt fresh by the wake login. Without it, a task could queue in kernel-SAM behind a device-sleeping LUN that never powers back up — an unrecoverable post-wake hang (immortal `SCSITaskUserClient`). HW-verified across 4 sleep cycles.
- **Docs/CI groundwork** (`6bf4fd1`): corrected panic-mechanism documentation and the `boot-args io=0` escape hatch. (This commit added the assert in the *absent* direction; `c853e78` flips it — kept as two commits for traceability.)

## Why the gate can go: the actual #54 mechanism

The gate existed because of the #54 boot panic, on the assumption that AMFI killing the dext at launch was the trigger. Desk research (fresh IOSCSIParallelFamily kernelcache disassembly + xnu source) says otherwise:

- A failed dext launch is harmless: the family kext contains **zero panic sites** (its failure path is a clean `false`-return unwind), and xnu's dext-death panics are compiled out for third-party dexts on release kernels. The reporter also had SIP disabled, so the dext most likely launched.
- The only reachable panic is xnu's **registry-quiesce watchdog**: if the FW643 PCI nub is still busy 60 s into boot, `watchdogd`'s `waitQuiet(…, kIOWaitQuietPanicOnFailure)` panics the machine (`IOService.cpp:5947`). The kernel shim drives HBA bring-up as synchronous calls with no timeout, so a dext that *blocks* anywhere in that chain converts to a boot panic.
- The old code had exactly such a block: the held-INQUIRY-against-target-0 model waited for an SBP-2 login that never came on an empty bus. **#55/#61 removed it** — the login-driven-target model publishes a target only when a device actually logs in, and its start chain never waits.

So the fix for #54 was never to avoid the PCI provider — it is a start chain that cannot stall. That property is what this PR pins down with the HW validation below and the CI assert.

## HW validation (M1 Pro / Tahoe 26 / FW643 via TB→FW / Nikon LS-9000)

All on one build (cdhash `18038de3…`, PCI-match personality, no mapper property anywhere):

- **The #54 scenario itself:** cold boot with adapter present, scanner off, 2+ min idle → no panic, HBA fully started (`busy 0 (1 ms)`), no target published.
- **Hot-plug bring-up without any device:** HBA start completes `busy 0 (3 ms)`.
- **Real scan:** login → target 0 created with real INQUIRY identity → VueScan preview with large READs. No mapper property needed — the PCI provider supplies the DART.
- **Sleep/wake ×2:** symmetric edges (`target 0 destroyed` at quiesce → `target 0 created` after wake), no kernel-side accumulation (1 target / 1 user client throughout), VueScan-restart recovery confirmed.
- **Unplug/re-plug under operation:** clean `Nub::Stop` → HBA `Stop` cascade, rebuild + re-login in ~0.3 s on re-plug.

Earlier sessions on the same model additionally covered teardown under load (in-flight 509 KB READ aborted cleanly) and scanner power-cycle re-login (no 5/26 regression).

## Caveats — and how to stage this if you prefer

- The #54 reporter's exact machine was never re-tested (the full `.panic` file would settle which stall variant fired, but nothing here depends on it).
- SIP-*on* launch behavior (expected: silent no-load) and Intel are unverified.
- The entitlement is not enforced under the self-signed model (SIP off + developer mode) the driver already requires — every HW run above was ad-hoc-signed with the entitlement embedded. And since distribution is one centrally signed zip with no local re-signing, default-off would mean SBP-2 users get nothing unless two zips ship.

If you'd rather keep SCSI opt-in for now, drop `c853e78` + `eefa9f5` — the mechanism analysis, the sleep fix, and the validation stand on their own.


mhellevang:
Pushed `c853e78` + `eefa9f5`: the `--scsi` gate is now removed in this PR (personality + entitlement in every build, CI asserts presence, single entitlements file, README rewritten to plain usage). Rationale and the honest residuals are in the updated body's last section — if you'd rather keep SCSI opt-in for now, just say so and I'll drop those commits; the bounded-start validation and the sleep fix stand on their own.

mrmidi:
Focused and clean, thanks.

Issue #100, opened 2026-08-17
Mackie Onyx 820i (Loud 0x000ff2): live capture — Oxford-run unit reports shared Onyx-i id 0x081216, enumerates cleanly on Tahoe


ottendorfcipher:
**Device:** Mackie Onyx 820i (Onyx-i series) — Oxford/OXFW971 production run
**Mac:** MacBook Pro M3 (Mac15,3), macOS 26.5.2 (Tahoe), ASFW 0.3.0 (local branch build)
**Chain:** Apple TB3→TB2 (A1790) → Apple TB2→FireWire 800 (A1463) → FW800 9-pin→FW400 6-pin cable → 820i

**Result:** enumerates cleanly — 2-node bus, mixer at S400 as bus root, full Config-ROM scan with zero failed transactions. Happy to attach the full 1394 diagnostics report if useful.

**Capture (from the ASFW app):**

- GUID: `0x000FF20400003AFC`
- Vendor ID: `0x000FF2` (Loud Technologies)
- Model ID: `0x081216`
- Unit directory: specifier `0x00A02D`, SW version `0x010001` (1394TA AV/C)
- ROM strings: `Loud Technologies Inc.` / `Onyx-i`

**Context / notes:**

- Mackie split the Onyx-i production run between Oxford OXFW971 ("former models", Linux `snd-oxfw`) and DICE ("latter models", Linux `snd-dice`, which gates Loud units on GUID category byte `0x10`).
- This unit's GUID category byte is `0x04` and it exposes a 1394TA AV/C unit → Oxford run. The model id matches libffado's `0x081216` "Onyx-i" OXFORD entry exactly.
- The DICE-run 820i/1220i/1620i model ids appear to be published nowhere (ALSA and libffado both match those units vendor-wide/generically), so a capture from a DICE-run unit would still be novel data.

I have a small recognition-only patch locally (Onyx family ids sourced from the libffado 2.5.0 device database, following the same recognized-but-not-audio-enabled policy as the PreSonus StudioLive siblings) — happy to open a PR if you'd like it.

Also interested in helping with the AV/C/Oxford audio path for these mixers — next step on my side would be an AV/C stream-format capture from the unit.

ottendorfcipher:
Follow-up: AV/C stream-format capture from the same Oxford-run 820i, done entirely through the ASFW MCP control plane (`asfw_fcp_send_command`, STATUS-class only — no writes to the device).

[...]

PR #101, opened 2026-08-17
Add Mackie Onyx-i (Oxford) support: recognition, publication, asymmetric duplex streaming

ottendorfcipher:
Adds support for the Mackie/LOUD Onyx-i mixers (Oxford OXFW971 production run). Developed and verified against an Onyx 820i behind an Apple TB3→TB2→FireWire adapter chain on macOS 26.5.2 / M3. Context and captures: #100.

## Verified on hardware

- Recognition + CoreAudio publication: 8-in / 2-out, 24-bit, 44.1 kHz, blocking mode.
- Capture: signal verified through the device's main-mix FW pair; sustained sessions with zero geometry mismatches and zero rejected packets.
- Playback: sustained music playback; steady-state TX (constant producer lag, no deficit growth, no short-frame faults).
- Duplex sessions start and stop cleanly (FSM terminal Idle, status 0).
- The device stamps usable SYT, so the existing RX-SYT-authoritative clock model works unmodified.

## Change summary (one commit per step)

1. `feat(profiles)` — recognition scaffold, sentinel-gated pending hardware capture.
2. `feat(profiles)` — recognition for the documented Onyx family ids from libffado 2.5.0 (Onyx-i Oxford `0x081216`, Onyx 1640i `0x001640`/`0x000006`, Blackbird `0x000007`); identity only, same policy as the StudioLive siblings.
3. `docs(profiles)` — live Config-ROM capture: GUID `0x000FF20400003AFC`, vendor `0x000FF2`, model `0x081216`, unit `0x00A02D`/`0x010001` (1394TA AV/C).
4. `docs(profiles)` — live unit-level `0xBF` stream-format capture: capture 8ch MBLA, playback 2ch MBLA, compound AM824, 44.1/48/88.2/96 kHz, one isoch plug per direction, no MIDI.
5. `feat(audio)` — publication (M1): asymmetric ADK isoch profile (TX 2ch/DBS 2, RX 8ch/DBS 8, blocking, 44.1 kHz) dispatched ahead of the DICE `GenericProfile()` catch-all; LOUD vendor-wide forced blocking per snd-oxfw `oxfw.c:189-196`.
6. `feat(avc)` — publication bypass: the OXFW971's Music subunit implements no descriptor mechanism (standard descriptor access returns NOT IMPLEMENTED, the non-standard direct read also refuses; the parse yields 0 plugs), so the generic descriptor-driven path cannot publish this device. Publishes the profile-owned configuration from `HandleInitializedUnit`, same pattern as the BeBoB bypass.
7. `feat(audio)` — streaming (M2): `MackieOnyxProtocol` runtime duplex control; Onyx joins the IRM any-channel mask and the BeBoB start order (no pre-stream clock lock — the device is SYT-unaware per snd-oxfw); LOUD wrong-DBS RX stride quirk plumbed profile → coordinator → transport → consumer → `RxAudioPacketProcessor` (ALSA `CIP_WRONG_DBS` semantics, `amdtp-stream.c:766-769`); header dbs remains visible in telemetry.
8. `fix(audio)` — start-clock pin: a first-ever start has no session clock and the coordinator's default requests 48 kHz, which a 44.1-only device refuses at Prepare (`kIOReturnUnsupported`). Adds an Onyx entry to `EffectiveStartClockForProfile`, same mechanism as the Duet's pin.
9. `fix(audio)` — `TxStreamPolicy.emptyPacketsDuringIdle = true`: record-only sessions otherwise collapse with a monotonically growing TX prep deficit (observed 3.3k → 5.7k frames before session teardown, five consecutive cycles). Matches Phase88's policy on the same base.

## Design decisions flagged for review

- **`MackieOnyxProtocol` subclasses `BeBoBProtocol`.** The base is protocol-neutral AV/C+CMP in practice: FCP dispatch, INPUT/OUTPUT PLUG SIGNAL FORMAT (`0x18`/`0x19` — the same command snd-oxfw uses for OXFW rate control), CMP connect/verify/teardown, plug-0 streams, no-op mixer hook. If a second OXFW device adopts it, renaming the base to something family-neutral would be the follow-up.
- **Rates are pinned to 44.1 kHz** (the device's captured current rate) until an AV/C rate transition is validated against this hardware. The base's `ApplyClockConfig` already speaks the right command, so this should largely reduce to widening the rate list later; 88.2/96 kHz additionally sit behind `AudioClockConfig`'s driver-wide gate.
- **The model id `0x081216` is shared across the Oxford-run Onyx-i family**; the static geometry here is 820i-verified. Siblings with different channel widths need a discovery-derived profile before they can stream (noted in the profile header; `RegisterBeBoBProfile` is the precedent).
- **Tests:** registry/factory integration-mode and identity pins; an asymmetry pin in `DiceProfileTests` locking 8-in/2-out against the Generic catch-all; an RX test pinning both the wrong-DBS quirk decode and the quirk-off failure mode (an oversized lying header decodes as zero events — indistinguishable from a silent device — rather than a geometry mismatch). The coordinator test mock is updated for the widened transport signatures.

References (Linux snd-oxfw/snd-dice/bebob, libffado 2.5.0, snd-firewire-ctl-services) were used as behavioral sources only, per the repository's reference policy; citations are in code comments. Local version-bump commits are excluded from this branch.

## Not included

- Sample rates beyond 44.1 kHz (see above).
- Two resilience issues observed during bring-up that are independent of this device support — summarized in #100, reproducible with logs available: idle Thunderbolt adapter drop/re-attach cycles, and the audio endpoint not re-arming after such a cycle (subsequent StartIO returns `kIOReturnNotReady` until a fresh boot). Happy to file both as separate issues.


ottendorfcipher:
Pushed three follow-up commits from continued hardware testing on the same 820i:

1. `6d70bd8` — safety-offset tuning: the Duet placeholder values left ~1.45 ms of TX headroom against observed host callback spikes of ~922 µs; widened per the measurements in the commit message.
2. `2ba3f97` — a 48 kHz rate-widening attempt (kept in history because its failure documented two engine findings).
3. `dd899ac` — its revert, after root-causing the regression it exposed: `AudioDuplexCoordinator::RequestClockConfig` persists `session.pendingClock` before executing the request, and the failure path (`FailPendingClockRequest` → `clockRequests_.FailPending`) completes the request bookkeeping without scrubbing the session. A rate change that fails at the ADK transport reconfig — which is currently unimplemented for AV/C static-profile devices (`HandleChangeSampleRate` → `transport reconfig failed: kIOReturnUnsupported`) — therefore leaves `pendingClock=48000` armed; the next session start consumes it and programs the device to 48 kHz via SIGNAL FORMAT while the host graph runs 44.1, producing periodic audible dropouts (`[TxExposure] reason=rate-mismatch`, ~7,000 ppm sawtooth). Details and the defense-in-depth rationale are in the commit message.

Net state of the branch: 44.1 kHz-only, validated full duplex on hardware (capture and playback, including sustained mic use). The rate widening should return together with (a) the pending-clock hygiene fix and (b) AV/C rate-change support in the ADK reconfig path.

I can file the pending-clock hygiene defect and the reconfig gap as separate issues (plus the two bring-up resilience observations already summarized in #100) if that is useful.

PR #102, opened 2026-08-29
fix(async): watchdog never starts, so transaction timeouts never fire

mhellevang:
## What's broken

The driver has an internal alarm clock (the async watchdog). Its one job: if a command never gets an answer, give up after a while and report an error.

Since `22c82112` (Aug 5, on main — also shipped in the 0.3.0 release), that clock is never started. The reason is a one-line startup-order bug: the clock refuses to start unless the driver is in the "running" state — but the clock is started *during* startup, before "running" is set (that commit changed the check; `ScheduleAsyncWatchdog` in `ASFWDriver.cpp`). The clock also only re-arms itself from its own tick, so if the first start is skipped, it never ticks at all. No error, no log line — it just silently isn't there.

## When it matters

As long as every device answers every command, the clock is never needed, so day-to-day use looks normal. It only matters when an answer goes missing — a sick device, a flaky cable, a lost interrupt. Then:

- the command waits **forever** instead of failing
- the app that sent it freezes (only force-quit helps)
- the driver's own recovery (reset, re-login) never runs
- each stuck command permanently eats a slice of the driver's resources — enough of them and FireWire stops working entirely until reinstall or reboot

## How we found it

A scanner's power supply died in the middle of a scan, in exactly the way where a timeout is the only possible rescue (the device stayed visible on the bus but stopped answering). VueScan froze permanently. Digging through the logs showed no timeout had fired in the entire session — and the code showed why.

## The fix

Three small changes:

1. Start the clock correctly during startup (allow starting while the driver is starting up, not just when fully running).
2. Make sure the clock can never accidentally stop itself again — a tick that has nothing to do still schedules the next tick. Shutdown stops it the proper way, as before.
3. One SBP-2 timer (management ORB) was started only *after* its command got a reply — so a command with no reply had no timer at all. It now starts before the command is sent.

## Testing

- All 1589 host tests green, plus a new test that fails on the old code.
- Real hardware (Nikon LS-9000, SBP-2): normal scanning works, and pulling the power mid-scan now gives the app an immediate error instead of a freeze.

### DIFF
--- a/ASFWDriver/Protocols/SBP2/SBP2ManagementORB.cpp
+++ b/ASFWDriver/Protocols/SBP2/SBP2ManagementORB.cpp
@@ -150,6 +150,24 @@ bool SBP2ManagementORB::Execute() noexcept {
 
     inProgress_.store(true, std::memory_order_relaxed);
 
+    // Arm the timeout BEFORE the write goes out so it covers the whole
+    // exchange (agent write + status block), not just the wait for status.
+    // Previously armed in OnWriteComplete: a write whose completion never
+    // arrived (wedged target — observed on LS-4000, LUN reset after ORB
+    // timeout) left the management ORB pending forever with no timer running.
+    if (scheduler_ != nullptr && timeoutMs_ > 0) {
+        const std::weak_ptr<int> weakLifetime = lifetimeToken_;
+        const uint64_t delayNs = static_cast<uint64_t>(timeoutMs_) * 1'000'000ULL;
+
+        timeoutToken_ = scheduler_->ScheduleAfter(delayNs, [this, weakLifetime]() {
+            if (weakLifetime.expired()) {
+                return;
+            }
+            timeoutToken_ = kInvalidSchedulerToken;
+            OnTimeout();
+        });
+    }
+
     // Write ORB address to management agent
     const FW::Generation gen{generation_};
     const FW::NodeId node{static_cast<uint8_t>(nodeID_ & 0x3Fu)};
@@ -172,6 +190,10 @@ bool SBP2ManagementORB::Execute() noexcept {
 
     if (!writeHandle_) {
         ASFW_LOG(Async, "SBP2ManagementORB::Execute: WriteBlock failed");
+        if (scheduler_ != nullptr && timeoutToken_ != kInvalidSchedulerToken) {
+            scheduler_->Cancel(timeoutToken_);
+            timeoutToken_ = kInvalidSchedulerToken;
+        }
         inProgress_.store(false, std::memory_order_relaxed);
         return false;
     }
@@ -193,22 +215,10 @@ void SBP2ManagementORB::OnWriteComplete(Async::AsyncStatus status,
         return;
     }
 
-    // Management agent write ACK'd. Start timeout, wait for status block.
+    // Management agent write ACK'd; the timeout armed in Execute() keeps
+    // running until the status block arrives.
     ASFW_LOG(Async, "SBP2ManagementORB: mgmt agent ACK'd, waiting for status block (timeout=%ums)",
              timeoutMs_);
-
-    if (scheduler_ != nullptr && timeoutMs_ > 0) {
-        const std::weak_ptr<int> weakLifetime = lifetimeToken_;
-        const uint64_t delayNs = static_cast<uint64_t>(timeoutMs_) * 1'000'000ULL;
-
-        timeoutToken_ = scheduler_->ScheduleAfter(delayNs, [this, weakLifetime]() {
-            if (weakLifetime.expired()) {
-                return;
-            }
-            timeoutToken_ = kInvalidSchedulerToken;
-            OnTimeout();
-        });
-    }
 }
 
 void SBP2ManagementORB::OnStatusBlockWrite(uint32_t offset,
--- a/ASFWDriver/Protocols/SBP2/SBP2ManagementORB.cpp
+++ b/ASFWDriver/Protocols/SBP2/SBP2ManagementORB.cpp
@@ -184,7 +184,14 @@ bool SBP2ManagementORB::Execute() noexcept {
         gen, node, mgmtAddr,
         std::span<const uint8_t>{orbAddressBE_.data(), orbAddressBE_.size()},
         speed,
-        [this](Async::AsyncStatus status, std::span<const uint8_t> response) {
+        [this, weak = std::weak_ptr<int>(lifetimeToken_)](
+            Async::AsyncStatus status, std::span<const uint8_t> response) {
+            // The timeout is armed before this write, so OnTimeout can fire —
+            // and Complete() can destroy the ORB — while the write is still
+            // in flight. Guard like the timer lambda above.
+            if (weak.expired()) {
+                return;
+            }
             OnWriteComplete(status, response);
         });
 
--- a/ASFWDriver/ASFWDriver.cpp
+++ b/ASFWDriver/ASFWDriver.cpp
@@ -930,7 +930,15 @@ void ASFWDriver::ScheduleAsyncWatchdog(uint64_t delayUsec) {
         return;
     }
     auto& ctx = *ivars->context;
-    if (!ctx.lifecycle || !ctx.lifecycle->AdmitsNormalWork()) {
+    // Gate on the alive states (kStarting/kRunning), NOT AdmitsNormalWork():
+    // the initial arm in StartRuntime happens before CompleteStart() flips the
+    // state to kRunning, so a kRunning-only gate silently no-ops it and the
+    // self-rearming tick chain never starts — no AT transaction timeout can
+    // ever fire (0.3.0 regression, 22c82112; observed as an unrecoverable
+    // SBP-2 wedge when a target stopped responding). Teardown still kills the
+    // chain authoritatively via watchdog.Stop()/Reset() (timer disable), and
+    // every return to kRunning re-enters StartRuntime, which re-arms.
+    if (!ctx.lifecycle || !ctx.lifecycle->AdmitsBringupInterrupts()) {
         return;
     }
     ctx.watchdog.Schedule(delayUsec);
@@ -942,12 +950,15 @@ void ASFWDriver::AsyncWatchdogTimerFired_Impl(ASFWDriver_AsyncWatchdogTimerFired
 
     if (ivars && ivars->context) {
         auto& ctx = *ivars->context;
-        if (!ctx.lifecycle || !ctx.lifecycle->AdmitsNormalWork()) {
-            return;
+        // Skip only the WORK when normal work is inadmissible — never the
+        // reschedule below. An early return here breaks the self-rearming
+        // chain permanently on a transient non-running state; the chain's
+        // real kill switch is the timer disable in watchdog.Stop()/Reset().
+        if (ctx.lifecycle && ctx.lifecycle->AdmitsNormalWork()) {
+            ctx.watchdog.HandleTick(ctx.controller.get(), ctx.deps.asyncController.get(),
+                                    ctx.isoch.ReceiveContext(), ctx.isoch.TransmitContext(),
+                                    ctx.statusPublisher);
         }
-        ctx.watchdog.HandleTick(ctx.controller.get(), ctx.deps.asyncController.get(),
-                                ctx.isoch.ReceiveContext(), ctx.isoch.TransmitContext(),
-                                ctx.statusPublisher);
     }
 
     ScheduleAsyncWatchdog(kAsyncWatchdogPeriodUsec);
--- a/tests/protocols/SBP2ORBTests.cpp
+++ b/tests/protocols/SBP2ORBTests.cpp
@@ -201,6 +201,41 @@ TEST(SBP2ORBTests, ManagementORBTimesOutWhenAgentWriteNeverCompletes) {
     EXPECT_EQ(-2, completionStatus);
 }
 
+TEST(SBP2ORBTests, ManagementORBTimeoutWithWriteInFlightThenLateWriteCompletion) {
+    // With the timeout armed before the agent write (Execute, not
+    // OnWriteComplete), OnTimeout can fire while the write is still in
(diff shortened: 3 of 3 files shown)


ASFWDriver/Protocols/SBP2/SBP2ManagementORB.cpp
```diff
@@ -150,6 +150,24 @@ bool SBP2ManagementORB::Execute() noexcept {
...
+        const uint64_t delayNs = static_cast<uint64_t>(timeoutMs_) * 1'000'000ULL;
+
+        timeoutToken_ = scheduler_->ScheduleAfter(delayNs, [this, weakLifetime]() {
+            if (weakLifetime.expired()) {
+                return;
+            }
+            timeoutToken_ = kInvalidSchedulerToken;
+            OnTimeout();
+        });
+    }
```
mrmidi, line 169: This relocation is the whole of the change — these 12 lines are byte-identical to the block deleted from `OnWriteComplete`, so a move-aware diff shows zero changed logic. But the new *position* removes an invariant that was never written down:

```
arm site inside OnWriteComplete  ⟹  the write has already completed when the timer starts
                                 ⟹  OnTimeout can never run while writeHandle_ is live
                                 ⟹  the ORB can never be destroyed with a pending write callback
```

Arming here, above `bus_.WriteBlock` at :183, deletes that chain. `OnTimeout` can now fire with the agent write still outstanding — which is fatal:

1. `OnTimeout` (:236) → `Complete(-2)` (:244).
2. `Complete` invokes the completion callback (:258), which is the one `CommandExecutor::SubmitTaskManagement` installed — and at `CommandExecutor.cpp:378` it does `managementORB_.reset()` on a `unique_ptr`. **The ORB deletes itself from inside its own method.** (The comment at :251-254 shows that self-destruction was already known about, and defended against for the `std::function` captures — just not for this.)
3. `~SBP2ManagementORB` (:27-34) cancels `timeoutToken_`, resets `lifetimeToken_`, and deallocates the address ranges — but **never touches `writeHandle_`**. It can't clean itself up either: `Async::AsyncHandle` is `FWHandle`, a bare `struct { uint32_t value; }` (`AsyncTypes.hpp:51-58`). No RAII, no cancel-on-destroy, and `IFireWireBus` exposes no cancel/abort at all.
4. Three callbacks capture `this`; the destructor neutralises two. The write lambda 18 lines below at **:187** captures raw `this` with no `lifetimeToken_` guard — unlike the timer lambda you just moved, 25 lines above it.
5. The AT layer resolves that transaction later — a genuinely late ACK, or `kIOReturnTimeout` from `TransactionCompletionHandler::OnTimeout` (`:307`) — and calls the lambda on freed memory.

Note the interaction: **commit 1 is what guarantees step 5 fires.** With the watchdog dead, that transaction sat there forever and never called back. The two commits are individually plausible and jointly lethal.

Confirmed under ASan — `heap-use-after-free`, write at `Complete:245` via `OnWriteComplete:214` via `Execute()::$_1:188`, freed by `unique_ptr::reset`. Clean on the pre-PR file. Recipe and the three-run before/after/fixed procedure are in the review summary.

Not just theoretical on hardware: the ORB timeout floors at **1000 ms** (`managementTimeoutMs + kManagementTimeoutGraceMs`, where `managementTimeoutMs = timeoutUnits * 500` straight from Unit_Characteristics with **no floor** — `SessionRegistry.cpp:49`; a target advertising `timeoutUnits = 0` gets exactly 1000). Meanwhile the AT write can stretch to **~950 ms** via the retry extensions in `TransactionCompletionHandler.hpp:517-583` (`HandleBusyTimeout` 3×200 ms, `HandleAwaitingARTimeout` 3×250 ms). Nothing enforces that ordering — it's an accident of two constants chosen independently in two subsystems.

**Fix, at line 187** — restores symmetry with the timer lambda:

```cpp
writeHandle_ = bus_.WriteBlock(
    gen, node, mgmtAddr,
    std::span<const uint8_t>{orbAddressBE_.data(), orbAddressBE_.size()},
    speed,
    [this, weak = std::weak_ptr<int>(lifetimeToken_)](
        Async::AsyncStatus status, std::span<const uint8_t> response) {
        if (weak.expired()) {
            return;
        }
        OnWriteComplete(status, response);
    });
```

Verified clean under ASan with all 11 `SBP2ORBTests` passing. Anchoring here rather than on :187 because that line isn't in the diff.

This makes the safety property *local*: `lifetimeToken_.reset()` in the destructor then disarms every captured-`this` callback, instead of correctness depending on timeout arithmetic in another subsystem. Cancelling `writeHandle_` would be stronger (it'd reclaim the tLabel too), but `IFireWireBus` has no cancel API — that's a follow-up, not a blocker here.

mrmidi (changes requested):
Reviewed against `main`. Verified locally: **1589/1589 host tests pass** on this branch, and the new regression test genuinely fails when I revert only `SBP2ManagementORB.cpp` — good test.

One blocker, in the SBP-2 commit. Details inline on `SBP2ManagementORB.cpp:169`; summary and validation steps below.

---

## Commit 1 (`221cbf72`, watchdog) — correct, no changes requested

The diagnosis is exactly right. `22c82112` swapped the gate `ctx.stopping` → `AdmitsNormalWork()` and left the arm at `ASFWDriver.cpp:424`, which runs before `CompleteStart()` at `:504`. Silent no-op, self-rearming chain, never starts.

Three things I checked that the PR doesn't spell out, all of which hold:

- **The widened gate costs nothing at bring-up.** The watchdog timer is created on the `"Default"` queue (`DriverContext.cpp:318-326` → `PrepareWatchdog:379`) — the same serial queue `StartRuntime` runs on. The first tick can't be delivered until `StartRuntime` returns, by which point the state is already `kRunning`. No no-op tick storm during startup.
- **Teardown still wins.** `BeginQuiesce` flips the state before `QuiesceRuntime` reaches `ctx.watchdog.Stop()`, and `ScheduleAsyncWatchdog`'s own gate rejects `kQuiescing`/`kStopped`. "Always reschedule" can't outlive teardown.
- **The re-arm invariant is real.** Every return to running goes through `StartRuntime` — wake at `:626`, recovery restart at `:699`.

Minor: the commit message's *"a transient non-running state can no longer break the chain permanently"* is stronger than the code delivers. The tick handler no longer breaks it, but `ScheduleAsyncWatchdog` still returns early for any state outside `kStarting`/`kRunning`, so a tick landing in `kSuspended` still ends the chain. Harmless — the load-bearing invariant is the `StartRuntime` re-arm, which the code comment does state correctly.

---

## Commit 2 (`09a7207e`, SBP-2) — use-after-free

Full mechanism in the inline comment. Short version: moving the arm above `bus_.WriteBlock` deletes an unwritten invariant — *the timer is armed ⟹ the write already completed* — so `OnTimeout` can now fire with `writeHandle_` live. `Complete()` → the `CommandExecutor` callback → `managementORB_.reset()` destroys the ORB from inside its own method, and the write lambda at `:187` captures raw `this` with no `lifetimeToken_` guard.

**Not asking you to revert this commit.** The relocation is defensible and the regression test is worth keeping. It needs one addition — the guard on the write lambda at `:187`, restoring symmetry with the timer lambda you moved.

To be fair on ownership: the missing guard predates this PR. The unguarded lambda, the destructor ignoring `writeHandle_`, and the self-destruct in `CommandExecutor.cpp:378` were all already there. This commit only makes them reachable.

One reframe worth making in the commit message: with commit 1 restored, **this commit is not what fixes the reported hang.** The AT write already carries its own 200 ms timeout (`Submitter.cpp:122`), and `TransactionCompletionHandler::OnTimeout` invokes the response handler with `kIOReturnTimeout` at `:307` — so `OnWriteComplete` → `Complete(-1)` fires roughly an order of magnitude before the ≥1000 ms ORB timeout. Commit 2 is defense-in-depth, and the message currently reads as though it's the cure.

---

## Validating this with ASan

The repo's `tests/` CMake project takes sanitizer flags at configure time, so this needs no changes to any build file. Use a **separate build dir** so `build/tests_build` stays clean:

```bash
cmake -S tests -B build/asan \
  -DCMAKE_BUILD_TYPE=Debug \
  -DCMAKE_CXX_FLAGS="-fsanitize=address -fno-omit-frame-pointer -g" \
  -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=address"

cmake --build build/asan --target SBP2ORBTests -- -j8
./build/asan/protocols/SBP2ORBTests
```

`-fno-omit-frame-pointer` is what makes the trace readable. Single-target builds keep the loop to a few seconds.

**Important: running ASan over the existing suite finds nothing here.** `ManagementORBTimesOutWhenAgentWriteNeverCompletes` stack-allocates the ORB, so it outlives everything and there is no lifetime error to detect. ASan is a detector, not a search engine — the test has to drive the *production* ownership pattern. This one mirrors `CommandExecutor::SubmitTaskManagement` exactly:

```cpp
TEST(SBP2ORBTests, ManagementORBTimeoutWithWriteInFlightThenLateWriteCompletion) {
    ORBTimerRig rig;

    auto orb = std::make_unique<SBP2ManagementORB>(rig.bus, rig.bus, rig.addressManager,
                                                   reinterpret_cast<void*>(0x8));
    orb->SetFunction(SBP2ManagementORB::Function::LogicalUnitReset);
    orb->SetLoginID(0x12);
    orb->SetManagementAgentOffset(0x80);
    orb->SetTargetNode(1, 0x3F);
    orb->SetTimeout(5);
    orb->SetScheduler(&rig.scheduler);

    int completionStatus = 99;
    orb->SetCompletionCallback([&](int status) {
        completionStatus = status;
        orb.reset();                  // what CommandExecutor.cpp:378 does
    });

    ASSERT_TRUE(orb->Execute());
    ASSERT_EQ(1u, rig.bus.PendingWriteCount());

    rig.AdvanceMs(5);                 // timeout fires; ORB destroys itself
    EXPECT_EQ(-2, completionStatus);
    EXPECT_EQ(nullptr, orb.get());

    // AT layer resolves the transaction afterwards — real late ACK, or
    // kIOReturnTimeout from the watchdog commit 1 just revived.
    EXPECT_TRUE(rig.bus.CompleteNextWrite(ASFW::Async::AsyncStatus::kTimeout));
}
```

`DeferredFireWireBus` holding writes until you explicitly complete them is what makes the window observable.

Result on this branch:

```
==76090==ERROR: AddressSanitizer: heap-use-after-free on address 0x611000000254
WRITE of size 1 at 0x611000000254 thread T0
    #2 SBP2ManagementORB::Complete(int)          SBP2ManagementORB.cpp:245
    #3 SBP2ManagementORB::OnWriteComplete(...)   SBP2ManagementORB.cpp:214
    #4 SBP2ManagementORB::Execute()::$_1         SBP2ManagementORB.cpp:188
   #11 DeferredFireWireBus::CompleteNextWrite    DeferredFireWireBus.hpp:131
0x611000000254 is located 212 bytes inside of 256-byte region [0x611000000180,0x611000000280)
freed by thread T0 here:
    #2 std::unique_ptr<SBP2ManagementORB>::reset   ← CommandExecutor.cpp:378
```

The write at offset 212 is `inProgress_.store(false)` on the freed object.

**Three runs pin the claim:**

1. **This branch, unfixed** → the ASan report above.
2. **`git checkout 221cbf72 -- ASFWDriver/Protocols/SBP2/SBP2ManagementORB.cpp`, rebuild** → test fails its *assertions* (status stays 99, ORB never destroyed) but **no ASan error**. So this commit introduces reachability.
3. **With the guard applied** → clean, 11/11 `SBP2ORBTests` pass.

Two caveats for this codebase: it only works for host tests (`ASFW_HOST_TEST` stubs DriverKit — you can't ASan a dext), and `detect_leaks` is unsupported on Darwin/arm64, so this is use-after-free / overflow / double-free only.

---

## Follow-up, not blocking this PR

Reviving the tick also revives what it carries, and it isn't really a watchdog — it's a 1 kHz composite poll on the Default queue: `OnTimeoutTick()` walks every live transaction and heap-allocates a `std::vector<TLabel>` per millisecond (`Tracking.hpp:296-327`, which has its own `// TODO: Optimize with priority queue/timer wheel`); plus isoch RX/TX `Poll()` that duplicates the interrupt path (`InterruptDispatcher.cpp:36-45`, `:67-75`) and only covers context 0, so it isn't even a complete backup on multi-stream devices; plus a `StatusPublisher::Publish()` every tick.

Not a reason to hold this PR — the dev branches have been running that tick alive the whole time; only `main`/0.3.0 has had it dead, so merging re-syncs `main` with what's actually been exercised.

The better shape already exists in-tree: `DriverKitSessionScheduler` is one `IOTimerDispatchSource` + `OSAction` re-armed at the earliest pending deadline (`:74-82`, `:133-228`). `IOTimerDispatchSource::WakeAtTime` is documented to replace an unfired deadline, so one timer suffices and the 1 kHz wake disappears. Separate PR.

ASFWDriver/Protocols/SBP2/SBP2ManagementORB.cpp
mrmidi: Correction to my comment above: I said `IFireWireBus` has no cancel API. That's wrong — `Cancel(AsyncHandle)` is declared on `IFireWireBusOps.hpp:147` and `IFireWireBus` inherits it (`FireWireBusImpl::Cancel` at `FireWireBusImpl.cpp:118` forwards to `async_.Cancel`). I had grepped only `IFireWireBus.hpp`, which is just a 19-line header combining Ops + Info.

So cancelling `writeHandle_` in `~SBP2ManagementORB` *is* available today, and it's the stronger fix — it reclaims the tLabel instead of leaving a zombie transaction to expire on its own.

This doesn't change the request: the `weak_ptr` guard is still the right change here. It's local, it makes the destructor disarm every captured-`this` callback uniformly, and it's safe regardless of what the AT layer does with an in-flight handle. Cancelling as well would be a reasonable belt-and-braces addition if you want it, but it needs its own thought about re-entrancy (`Cancel` may invoke the completion synchronously, from inside the destructor). Guard first; cancel is optional and separable.

mrmidi (approved):

PR #106, opened 2026-09-13
Add experimental Focusrite Saffire Pro 40 support

evan-luther:
## Summary

Enables experimental audio support for the original Focusrite Saffire Pro 40 (`0x00130e:0x000005`, TCD2220) without enabling the distinct TCD3070 revision.

- Publishes the model through the existing DICE/TCAT path.
- Adds per-stream playback geometry so asymmetric DICE devices can select disjoint CoreAudio channel slices.
- Configures the Pro 40 as 12 PCM + 1 MIDI and 8 PCM playback streams, plus two 10 PCM + 1 MIDI capture streams: 20 inputs and 20 outputs in CoreAudio.
- Keeps the original Pro 40 on blocking AM824 streams at 44.1/48 kHz.
- Updates compatibility documentation and the changelog.

## Wire/reference basis

The geometry was captured from Focusrite GUID `0x00130E0401405B54`, TCAT product/model `0x005`, TCD2220. It was cross-checked against FFADO `src/dice/focusrite/saffire_pro40.cpp:50-97`, whose router map exposes ARX0 12 + ARX1 8 host-to-device channels and ATX0 10 + ATX1/loop 10 device-to-host channels. Blocking cadence matches Linux `sound/firewire/dice/dice-stream.c:488-514` (`CIP_BLOCKING`). Reference code was not copied.

The TCD3070 IDs (`0x0000de` and GUID board-model `0x13`) remain recognized but disabled because this hardware was not tested and Linux documents a different single-stream layout.

## Validation

- `./build.sh --test-only`: **1,592 tests, 0 failures** (six repository fixtures skipped).
- `./build.sh --test-only --test-filter 'DiceProfileTests|AudioProfileRegistryTests|DeviceProtocolFactoryTests'`: **41 focused tests, 0 failures**.
- `DEVELOPER_DIR=/Applications/Xcode-beta.app/Contents/Developer ./build.sh --no-bump`: Debug app/dext build succeeded. Existing Xcode 27 beta Swift-concurrency warnings remain outside this change.

## Hardware results

One original Saffire Pro 40 was exercised at 48 kHz on both machines through an adapter chain exposing the Agere `pci11c1,5901` FireWire controller:

- MacBook Pro `Mac16,8` (M4 Pro), macOS 27.0: 20-in/20-out CoreAudio publication; audible playback; an eight-second capture probe returned 3,066,878 nonzero samples with input 1 peak `0.0889`; Ableton playback/capture confirmed.
[...]

PR #114, opened 2026-09-14
MOTU UltraLite (protocol v2) device support

deweydb:
## MOTU UltraLite (protocol v2) support

Rebased onto `main` and split, as asked. This is now **device support only**.

Adds the MOTU protocol-v2 family (UltraLite first, 828mk2 enabled on the same tables) to the audio stack: packet framing, timing, device bring-up, channel naming and the host channel map.

Credit to **Jonathan Woodward** ([Dreambrother7/ASFireWire@feature/motu-wire-codec](https://github.com/Dreambrother7/ASFireWire/tree/feature/motu-wire-codec)) — his wire/codec layer is carried over, forward-ported onto current `main`.

### What changed since your review

- **Rebased onto `main`.** The Saffire Pro 40 revert is gone, and the six commits already merged via #108-#113 are gone with it.
- **Cycle master dropped** — your version in #117 is better: it re-arms per generation, mine only set a default bit.
- **CSR fix dropped** — you already cherry-picked it.
- **IRM role change split out** into #118 — please read that one: I think one of your premises was wrong, and the code says so.
- **Controls subsystem split out** into #119 (volume, mute, knob tracking), stacked on this PR.

### Dependency, stated plainly

**This PR alone will not stream on `main`.** The device enumerates, publishes, names its channels and builds its graph, but `ReservePlayback` fails with `kIOReturnNoDevice`, because a two-node bus whose peer does not contend has no IRM and no owner for `CHANNELS_AVAILABLE`. That is what #118 fixes. Merging this one on its own is harmless, just not yet audible.

### What it contains

- **Wire framing** (`Audio/Wire/MOTU/`): 3-byte PCM chunks at byte offset 10 behind the SPH quadlet; `MotuPayloadWriter`/`MotuPayloadReader`; `MotuEventOffsetCache`, `MotuRxTiming`, `MotuTxTiming` replay the device's per-block SPH presentation times, as Linux `amdtp-motu.c` does.
- **CIP quirks**: the header DBS is wrong so geometry comes from the chunk count (`CIP_WRONG_DBS`); transmit DBC is the end-event count (`CIP_DBC_IS_END_EVENT`); SYT is `0xFFFF` (`CIP_UNAWARE_SYT`); receive timing is established from SPH, not the SYT cadence.
- **Bring-up**: `MotuV2Protocol` implements `IDuplexDeviceControl`; `PrepareDuplex` applies the clock rate; the nub is published at discovery, which breaks a caps-versus-streaming deadlock (live caps only exist once streaming starts, and CoreAudio only streams a published device).
- **Identity**: MOTU publishes `model_id` 0, so models match on `Unit_Spec_Id` + `Unit_Sw_Version`, as Linux `motu.c` does. The device shows as "MOTU UltraLite".
- **Channel map** (`MotuPortLayout.hpp`): wire order puts the headphones first on output and the CueMix return first on input, so a plain mapping sends the Mac's default pair to the headphones and records the mixer. The map is a host-side permutation from FFADO's `PortGroups_ULTRALITE` — the same reordering FFADO does with `port_order`. Nothing on the wire changes.

| Host channel | Output | Input |
|---|---|---|
| 1-2 | Main L/R | Mic 1/2 |
| 3-10 | Analog 1-8 | Analog 3-8, S/PDIF 1-2 |
| 11-12 | S/PDIF 1-2 | Mix Return L/R |
| 13-14 | Phones L/R | (padding) |

### Shared code it touches, and why

All of it is gated on MOTU or defaulted so other families are unaffected:

- `DeviceProtocolFactory` / `AudioRuntimeRegistry` / `DeviceRegistry`: a defaulted unit-identity parameter, needed because `model_id` cannot identify this family.
- `ASFWAudioNub::StartAudioStreaming`: passes that identity, otherwise a register-based family falls into the AV/C rebind gate it can never pass.
- `ASFWAudioDriverGraph`: keeps the nub-supplied device name when it has one, instead of renaming an UltraLite to "Generic DICE".
- `DuplexStreamProfile` / `AudioDuplexCoordinator` / `IsochDuplexHostTransport`: carry the MOTU chunk count and port map.
- `RxAudioPacketProcessor`: a MOTU decode branch; `DiceTxStreamEngine`: selects the MOTU writer and stamps SPH; `AmdtpTxPacketizer`: an opt-in `dbcIsEndEvent` flag.

### Testing

- 1760/1760 host tests pass; Release dext builds `x86_64 arm64e`.
- Hardware: MOTU UltraLite, 48 kHz, macOS 26 on Apple Silicon, FW643 via a Thunderbolt Display. Playback on the main outs, capture from both mic inputs, and channel names confirmed by ear and in Audio MIDI Setup (with the IRM PR applied).
- Not tested: 44.1 kHz; the 828mk2 (same chunk geometry and port table, but I do not own one). 896HD/Traveler/8pre are named but left disabled. UltraLite-mk3 and later are protocol v3 and unsupported.

### Known issue, unrelated to this code

After reinstalling the dext the old DriverKit server sometimes lingers with no services, so the new build will not load until a reboot or `pkill`. I did not find the root cause.


mrmidi:
I've cherry-picked `11885caa` (zero the CSR data registers before a local IRM resource read) into #117 — it's device-independent, I'd verified the premise against handle_local_lock, and the stale-compare-swap corrupting `CHANNELS_AVAILABLE_31_0` seems a real bug on hardware that has nothing to do with MOTU.

For the rest, three things before it can land:

It currently reverts #106 (Saffire Pro 40) — it branched before that merged. Six of the 36 commits are also already in main via #108–#113.

The cycle-master change is now redundant: #117 carries a version that fixes it in `CyclePolicyCoordinator` and re-arms per generation. `kDefaultLinkControl` alone misses Linux's per-self-ID re-arm at `ohci.c`.

On `MakeLiveDefault() → IRMResourceHost:` the comment cites Linux setting `PHY_LINK_ACTIVE | PHY_CONTENDER`, but that's `ohci.c:2511` — a write to PHY register 4. This changes the Config ROM irmc bit instead, and that bit isn't what's arbitrated. I have a device on my bench that proves it: the Apogee Duet publishes IRMC=0 in its BIB and is nevertheless the designated IRM, because its Self-ID asserts contender + link-active. Raising our ROM irmc can't win an election that reads Self-ID, and kPhyContender (IEEE1394.hpp:64) is never written anywhere in the tree.

The change does do something real, though — `RoleMode::IRMResourceHost` also gates our own `IRMFallbackCoordinator`, `CyclePolicyCoordinator`, `GapPolicyCoordinator` and `RootSelectionCoordinator`. I suspect that's the effect you were after. If so it's worth separating: opening those gates is a policy decision I'd take on its own terms, and asserting PHY contender is a genuine gap but wire-visible and wants hardware in hand.

Could you rebase onto `main` and split? MOTU device support on its own, controls subsystem separately. The device work itself is good and I want it.

deweydb:
Rebased and split as asked. This PR is now device support only, and there are two follow-ups: #118 (IRM contender) and #119 (controls subsystem, draft, stacked on this one).

**Saffire revert and the already-merged commits** — gone with the rebase onto `main`.

**Cycle master** — dropped. Yours in #117 is the better fix; mine only set a default bit and missed the per-generation re-arm.

**CSR zeroing** — dropped, since you cherry-picked it.

**The IRM change** — split out into #118, and worth a look, because I think one premise there is wrong.

You wrote that `kPhyContender` is never written anywhere in the tree. It is, in `ControllerCoreLifecycle.cpp:131-149`, gated on the role:

```c++
const bool shouldAdvertiseContender =
    (policy.roleMode == FullBusManager && policy.fullBMActivityLevel >= ElectionOnly) ||
    policy.roleMode == IRMResourceHost ||          // <-- what MakeLiveDefault() selects in #118
    (policy.roleMode == LegacyBmcCleared && config.allowCycleMasterEligibility);

const uint8_t phyReg4Bits = shouldAdvertiseContender
    ? (kPhyLinkActive | kPhyContender)
    : kPhyLinkActive;
```

and when it is false, `kPhyContender` is explicitly cleared. So the one-line role change was never really about the BIB `irmc` bit you objected to — you were right that the ROM bit cannot win an election, and right that my comment cited Linux's PHY write for something that did not make one. What it actually does is assert the PHY contender bit, which is the gap you identified. #118 rewrites the rationale accordingly and corrects the stale comment above that block, which still claims the live driver is seeded with `FullBusManager`/`ForceRootAllowed`.

You were also right that this is wire-visible and wants hardware. The open question is in #118: on a two-node bus the host usually holds the higher physical ID, so your Duet may stop being the IRM and the host may win instead, moving resource hosting from remote locks to the local CSR engine. That is what Linux does on every bus, but it is a real change on hardware I cannot test, and I would rather you confirm it on your bench than take my word.

One thing to know before merging this PR on its own: **it will not stream without #118.** The device enumerates, publishes, names its channels and builds its graph, but `ReservePlayback` fails with `kIOReturnNoDevice`, because a bus whose peer does not contend has no IRM to own `CHANNELS_AVAILABLE`. Harmless to merge, just not yet audible.

Thanks for the review, and for taking the six small ones.

PR #118, opened 2026-09-16
bus: contend for IRM in the live role, so an isochronous host can win it

deweydb:
## Make the live role a contender, so a bus with no other contender still gets an IRM

Split out of #114 at your request, and rewritten: the rationale in that version was wrong, and the corrected one is narrower than it looked.

### What it changes

One line: `ControllerConfig::MakeLiveDefault()` returns `RoleMode::IRMResourceHost` instead of the struct default (`ClientOnly`). Plus the comment explaining why, and the capability tests.

### Why it works — the PHY contender bit, not the BIB bit

You were right that the BIB `irmc` bit is not what's arbitrated, and that my original comment cited Linux's PHY write for a change that didn't make one. Following that up, the tree does write it:

`ConfigurePhyOperationalRegisters()` (`ControllerCoreLifecycle.cpp:131-149`) gates PHY register 4 on the role:

```c++
const bool shouldAdvertiseContender =
    (policy.roleMode == FullBusManager && policy.fullBMActivityLevel >= ElectionOnly) ||
    policy.roleMode == IRMResourceHost ||          // <-- this PR selects this
    (policy.roleMode == LegacyBmcCleared && config.allowCycleMasterEligibility);

const uint8_t phyReg4Bits = shouldAdvertiseContender
    ? (kPhyLinkActive | kPhyContender)
    : kPhyLinkActive;
// ...and when false, kPhyContender is explicitly cleared.
```

So `kPhyContender` **is** written, and `ClientOnly` actively clears it. That is the whole mechanism: at `ClientOnly` the host is not a contender, so it can never be elected IRM by Self-ID, whatever its ROM says. Raising the role asserts the bit, and everything else follows from the ordinary election.

The comment directly above that code still says a live driver is seeded with `FullBusManager`/`ForceRootAllowed`, which stopped being true when `MakeLiveDefault()` went back to `RolePolicy{}`. This PR corrects it.

### What it fixes

On a two-node bus whose peer does not contend — a MOTU UltraLite is one — nobody is IRM. `CHANNELS_AVAILABLE`/`BANDWIDTH_AVAILABLE` have no owner, so every `StartIO` died in `ReservePlayback` with `AllocationStatus::NotFound` → `kIOReturnNoDevice`:

```
[Topology] gen=1 nodes=2 root=1 IRM=none local=1
[Hardware] FW-22: roleMode=1 advertising bmc=0 irmc=0 cmc=0 isc=0
[FSM] Failed class=StageFailure cause=ReservePlayback status=0xe00002c0
```

With the bit asserted the host wins unopposed, `LocalIRMResourceController` sees `localNodeId == irmNodeId`, and the OHCI CSR engine hosts the registers. Channel and bandwidth allocation then succeed and the device streams.

### Why it should be safe on a bus that already has an IRM

This does not make the host behave like a bus manager. Every mutating path stays shut unless we are actually the elected IRM *and* the fallback gate is open:

- `IRMFallbackCoordinator::OnTopologyReady` returns early unless `localIsIRM`.
- `GapPolicyCoordinator::IsAllowedActor` requires `localIsIRM && irmFallbackGateOpen && irmFallbackNoBMDetected`.
- `RootSelectionCoordinator::IsAllowedActor` requires the same conjunction.
- No BM election: that is `FullBusManager` only.

So where another node is the IRM, the only difference is our own Self-ID posture. Both reference stacks take that posture unconditionally: Linux sets `PHY_LINK_ACTIVE | PHY_CONTENDER` in `ohci_enable()` (`ohci.c:2510-2511`) and clears it only on card removal (`core-card.c:794`).

### Open question — worth your Duet before this lands

The election is decided by physical ID among contenders, and on a two-node bus the host is usually root with the higher ID. Your Duet contends and currently wins IRM on your bench. With this change the **host** may win it instead, which moves resource hosting from remote lock transactions to the local OHCI CSR engine.

That should be fine — it is what Linux does on every bus — but it is a real change on hardware I can't test, and it is the one thing in this PR I would want your Duet to confirm before merge. If it does regress, the alternative is to gate the contender assertion on "no other contender seen in Self-ID", which is uglier and races with the election it is trying to influence.

### Testing

- `CapabilityModeTests` updated: the live profile advertises IRM capability, and a dedicated case pins `ClientOnly`'s passive semantics so that mode keeps its meaning.
- Full host suite green.
- Hardware: MOTU UltraLite on macOS 26 (Apple Silicon, FW643 via Thunderbolt Display) — with this change IRM election succeeds and channel/bandwidth allocation works; without it, allocation fails every time.


mrmidi:
Thanks. What I took away from reading the IOFireWireFamily source: Apple's arbitration only kicks in when something asks for it — a DelegateCycleMaster property, a known-bad IRM, or a node pinned off root. Left alone it does nothing at all; the contender bit stays clear. Once something does trip it, though, it claims rather than defers ("Mac which wants to be cyclemaster gets priority" https://github.com/apple-oss-distributions/IOFireWireFamily/blob/ad9f879946570c786a80c731c8e9baaa490994ed/IOFireWireFamily.kmodproj/IOFireWireController.cpp#L2423). 

So Linux always sets the contender bit, Apple dodges bus-manager and cycle-master work until something forces it in. Both approaches are valid.

Without your fix, a bus where no device contends for IRM has no IRM at all — so nothing owns `CHANNELS_AVAILABLE` and we can't allocate an isochronous channel. That's real and I have no hardware to reproduce it; everything I own contends and wins IRM itself. Good to leave your fix in for now.

Issue #115, opened 2026-09-14
PreSonus StudioLive 24.4.2 – ASFW hardware capture


Annavtechno:
Hi!
this is Anna, from Greece, and I'm a happy owner of a PreSonus StudioLive 24.4.2. 

I really wanted to help with getting this beautiful mixer working again on modern Macs, so I followed the instructions and did the requested hardware capture.

### Hardware

Device: PreSonus StudioLive 24.4.2
Mac: Mac mini M1 (2020), 16 GB RAM
macOS: Tahoe 26.6.2
ASFW version: 0.3.0

FireWire chain:

Mac mini M1
→ Apple Thunderbolt 3 to Thunderbolt 2 Adapter
→ Apple Thunderbolt 2 to FireWire 800 Adapter
→ FireWire 800-to-400 cable
→ StudioLive 24.4.2

### Result

The StudioLive 24.4.2 was successfully discovered by ASFW.

Config ROM was read successfully and identified as:

PreSonus / STUDIOLIVE_2442
GUID: 0x000A9204049204CB

The DICE report was also successfully read.

The device was locked at 48 kHz.

### DICE stream layout

TX (device → host):
- Stream 0: 16 PCM channels (Ch 1–16)
- Stream 1: 16 PCM channels (Ch 17–24, Auxiliary Ch 25–32)

RX (host → device):
- Stream 0: 16 PCM channels (Ch 1–16)
- Stream 1: 10 PCM channels (Ch 17–24, 2TrackIn L/R)

Both TX and RX report 2 streams.

Audio telemetry did not become active, as expected, since the StudioLive 24.4.2 is not currently audio-enabled in ASFW.

The 1394 Diagnostics and DICE reports are attached.

After completing the capture, I uninstalled ASFW and re-enabled SIP on my Mac.

I hope this information helps you add full audio support for the StudioLive 24.4.2.

Thank you SO much for working on this project and for giving these FireWire devices a chance to live again on modern Macs.

And thank you especially for letting us owners contribute the hardware information you need. ❤️

Greetings from Athens

[ASFW_DICE_Report.txt](https://github.com/user-attachments/files/32202486/ASFW_DICE_Report.txt)
[ASFW_1394_Diagnostics_Report.txt](https://github.com/user-attachments/files/32202482/ASFW_1394_Diagnostics_Report.txt)

Bests,

Anna

mrmidi:
Hi Anna. Thanks on providing 24.4.2 profile.

[...]

PR #122, opened 2026-09-18
Enable PreSonus StudioLive 24.4.2 from its captured geometry

mrmidi:
Enables audio for the PreSonus StudioLive 24.4.2, using the register dump Anna contributed in #115.

The model was already recognised by name but deliberately left disabled, because its channel counts differ from the 16.0.2 and nobody had captured them. Her dump answers that:

```
capture  (device -> host): 16 ch + 16 ch  = 32
playback (host -> device): 16 ch + 10 ch  = 26
```

MIDI is 0 in every stream, so DBS equals the channel count. The 16.0.2 muxes one MIDI slot and runs DBS 17, which is why its profile couldn't just be pointed at another model id.

This is the first device with an asymmetric playback side, so the profile overrides two defaults that assume uniform streams — otherwise playback would be described as 16+16, which puts 16-slot CIP into a device stream that has 10, and CoreAudio would advertise six outputs that don't exist.

No custom start order. Every device that carries one is either not a plain DICE device (MOTU, Duet, BeBoB, both Onyx models) or has inverted clock mastership (Weiss). This one is bracketed by two devices already running the DICE defaults — the 16.0.2 for PreSonus, the Venice F32 for two streams per direction — and its dump reports `STATUS locked=yes`, so the pre-stream clock-lock gate has real state to read.

No quirks either: `snd-dice`'s PreSonus handling covers model `0x000008` only, and libffado lists a plain `driver = "DICE"` with no mixer entry, which matches the absent EAP space in the dump.

The 16.4.2 and 32.4.2 stay gated — still no capture for them. I left a note where their gate is: libffado records `xmit_transfer_delay = 4` for the 16.4.2, the only StudioLive with any quirk at all, and no register dump will ever show that.

### Testing

1814/1814 host tests pass, and the geometry is pinned by a new registry test.

But this has never run against the hardware. It's derived from a register dump and two reference stacks, and the asymmetric playback path has no prior user — Venice is 16/16 both ways, so that code has only ever been exercised with identical streams.

@Annavtechno — if you're willing to try this, I'd be very grateful, though I know it means reinstalling and turning SIP off again on a mixer you actually use, so please don't feel obliged. If it does misbehave, the most likely shape is the first 16 output channels working and the last 10 not, since those are the ones on the stream this PR treats differently.


mrmidi:
NB. DICE reports I've saw give us enough information to build proper geometry from capabilities DICE is announcing. We don't want separate profile for every single device for it probably.

PR #123, opened 2026-09-18
Run the host tests under ASan/UBSan, and fix the six defects it found

mrmidi:
Adds a sanitized build for the host tests, and fixes the six defects it found the first time it ran.

The recurring failure mode in this driver is lifetime — cross-service use-after-free on teardown, refcount imbalance, borrows that outlive what they borrow from. A plain test build can't see any of it: a use-after-free reads plausible bytes and the test passes. There was no sanitizer configuration in the project at all.

Turning it on found six real `stack-use-after-scope` defects already sitting in the suite.

### The six

Four are test-ordering mistakes — an object declared before the locals its destructor's callbacks write into, so destruction order kills the captures first:

- `EfcTransportTest.IgnoresResponsesFromOtherNodes`
- `FireworksProtocolTest.PrepareDuplexProbesHwInfoFirstWhenInitializeDidNotRun`
- `CompletionRefactorPlan.ARResponseRejectsDifferentNodeNumber`
- `CompletionRefactorPlan.BusyAckExtendsDeadlineNoCompletion`

The other two are more interesting. `FetchAgent` tracks outstanding ORBs by raw pointer and writes `SetAppended(false)` into each one when cleared, so a cleared ORB has to still be alive. Production does that properly — `CommandExecutor::CleanupCommandResources()` calls `ClearCommandTracking()` *before* dropping `commandORB_`, and `SessionRecord` declares its executor last so it tears down before the session. The tests were building bare stack ORBs with no executor doing that for them. They now perform the same deregistration.

Applied to all five tests that build ORBs, not just the two that failed — the others merely had no outstanding ORBs at teardown.

### Wiring

CI gets a sanitized job; `build.sh` gets `--asan`.

Both set `ASAN_OPTIONS=abort_on_error=1`, which matters more than it looks: without it ASan prints its report and the process still exits 0, so ctest calls it a pass. That's exactly how my first sanitized run under-reported its own results.

The CI job runs at half the cores. Each sanitized process reserves a large shadow mapping, and at full parallelism one test was seen aborting under memory pressure with no sanitizer report behind it — not reproducible since, at `-j4` or `-j8`, alone or in the full suite. If it shows up in CI it's real and worth chasing then.

### Testing

Sanitized and plain are both green. Test/build only — no driver code is touched, so the risk here is that CI gets slower, not that anything breaks.

Worth landing before the other branches I have queued, so they're validated by it rather than arriving alongside it.


mrmidi:
Pushed fixes for everything CI surfaced.

**The sanitizer finding was real, and not in the tests.** `AVCMusicCapabilitiesWire` declared its two padding bytes *after* its only 32-bit field, so `supportedRatesMask` sat at offset 2 — permanently misaligned — while the comment on that padding claimed it aligned the struct to 8 bytes. Both the driver and the Swift app cast raw buffers to this struct, so both were doing misaligned 32-bit accesses. It works on arm64 because unaligned loads are tolerated, which is why it went unnoticed.

Moving the padding ahead of the field puts the mask at offset 4 and leaves every other offset and the total size unchanged. `DriverConnectorModels.swift` parses this header by fixed byte offsets and is updated in lockstep (`2..<6` → `4..<8`); I verified by printing `offsetof` for every field Swift indexes — currentRate 1, mask 4, ports 8, numPlugs 14, sizeof 18, all matching. Two `static_assert`s now pin the alignment and the size, since nothing else connects the two languages.

**The six `[[nodiscard]]` warnings needed two different fixes.** Four discarded `gate_.Acquire()`, which returns whether the claim succeeded and returns false both when the guid is already claimed and when the lock is null — every one was setup for a following assertion, so a silent false meant the precondition was never established and the assertion passed anyway. Those are now `ASSERT_TRUE`. The other two discarded `UpsertFromROM()`, which returns data rather than status, so `(void)` is right there. Blanket-`(void)`-ing all six would have silenced the four that mattered.

Also worth recording: I originally wrote this finding off as a flaky abort under parallelism. It wasn't — UBSan defaults to print-and-continue, so my local runs without `UBSAN_OPTIONS` reported it as passing, and I compared runs with different instrumentation and called the difference flakiness. The `halt_on_error` settings in this PR are what make it deterministic.

